PR #16, opened 2025-05-15
Create CI pipelines

justinchuby:


.github/workflows/main.yml
```diff
@@ -0,0 +1,96 @@
...
+          python -m pip install -r requirements-dev.txt
+      - name: Versions
+        run: |
+          pip list | grep numpy
+          pip list | grep onnx
+          pip list | grep torch
+      - name: Install package
+        run: pip install .
+      - name: Build documentation
+        run: python -m sphinx docs dist/html
```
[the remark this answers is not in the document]
justinchuby: Will handle later

.github/workflows/main.yml
```diff
@@ -0,0 +1,96 @@
...
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() }}
+        uses: codecov/test-results-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+
+  build_docs:
```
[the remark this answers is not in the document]
justinchuby: Will handle later

.github/workflows/main.yml
```diff
@@ -0,0 +1,96 @@
+name: CI
+
+on:
+  schedule:
+    # Run weekly on Mondays and Wednesdays 00:00
+    - cron: '00 00 * * MON,WED'
```
titaiwangms, line 6: Why does CI need a cron?

.github/workflows/main.yml
```diff
@@ -0,0 +1,96 @@
...
+  push:
+    branches:
+      - main
+      - 'gh/**/base' # ghstack base branches
```
titaiwangms, line 10: Wow

.github/workflows/main.yml
titaiwangms, line 96: It seems there is no test.

titaiwangms (approved):

.github/workflows/main.yml
justinchuby: In onnxscript we wanted to periodically run the ci with the latest torch nightly. Here we test with torch nightly as well so maybe a good idea to keep the cron triggers. But we may not need it as frequent.

.github/workflows/main.yml
justinchuby: Contributors with write access can use ghstack

.github/workflows/main.yml
justinchuby: Thanks - trying to fix now

.github/workflows/main.yml
justinchuby: Reduced to weekly

PR #17, opened 2025-05-16
Create package description

justinchuby:


pyproject.toml
```diff
@@ -5,6 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "onnx-ir"
 dynamic = ["version"]
+description = "Efficient in-memory representation for ONNX"
```
titaiwangms, line 8: Should we include I (intermediate) from IR?

titaiwangms (approved):

pyproject.toml
justinchuby, line 8:
```suggestion
description = "Efficient in-memory intermediate representation for ONNX"
```

pyproject.toml
```diff
@@ -5,6 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "onnx-ir"
 dynamic = ["version"]
+description = "Efficient in-memory intermediate representation for ONNX"
```
gramalingam, line 8: I don't mind either way. But, technically, ONNX is itself "intermediate", when going from pytorch to backends (like ORT, tensor-rt). But this repo is about representing ONNX as it is, enabling round-tripping back to ONNX in most cases. So, "in-memory representation" is arguable more accurate. (Again, viewed from onnxscript perspective, it is an intermediate-representation when going from python/onnxscript, but that is the nature of ONNX itself). IMHO, it is not worth over-thinking unless someone complains.

justinchuby: I agree with the above. Had the same thoughts. I will revert to the original and merge.

justinchuby, line 8:
```suggestion
description = "Efficient in-memory representation for ONNX"
```

PR #33, opened 2025-05-29
Create `add` methods on initializers and attributes

justinchuby:
1. It's now possible to do `node.attributes.add(ir.Attr(...))` to avoid having to get the name as a redundant step.
2. Also allow node attributes to be initialized with a dictionary
    - This way users do not have to use
    ```py
    ir.Node(attributes=node.attributes.values())
    ```
	when copying the node. Instead, it is possible to

    ```py
    ir.Node(attributes=node.attributes)
    ```
3. Implement get_* methods on attributes to make getting default attributes easier.

    Before
    ```py
    attr = node.attributes.get("attr")
    if attr is not None:
        value = attr.as_int()
    else:
        value = 42
    ```

    now

    ```py
    value = node.attributes.get_int("attr", 42)
    ```

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -28,7 +28,6 @@ from collections.abc import (
     Iterable,
     Iterator,
     Mapping,
-    MutableMapping,
     MutableSequence,
     Sequence,
 )
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -22,12 +22,12 @@ import os
 import sys
 import textwrap
 import typing
-from collections import OrderedDict
 from collections.abc import (
     Collection,
     Hashable,
     Iterable,
     Iterator,
+    Mapping,
     MutableMapping,
     MutableSequence,
     Sequence,
@@ -39,7 +39,6 @@ from typing import (
     Any,
     Callable,
     Generic,
-    Mapping,
     NamedTuple,
     SupportsInt,
     Union,
@@ -1372,7 +1371,11 @@ class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
         self._inputs: tuple[Value | None, ...] = tuple(inputs)
         # Values belong to their defining nodes. The values list is immutable
         self._outputs: tuple[Value, ...] = self._create_outputs(num_outputs, outputs)
-        self._attributes: _graph_containers.Attributes = _graph_containers.Attributes(self, attributes)
+        if isinstance(attributes, Mapping):
+            attributes = tuple(attributes.values())
+        self._attributes: _graph_containers.Attributes = _graph_containers.Attributes(
+            attributes
+        )
         self._overload: str = overload
         # TODO(justinchuby): Potentially support a version range
         self._version: int | None = version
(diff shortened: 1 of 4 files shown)


src/onnx_ir/_graph_containers.py
```diff
@@ -266,3 +266,26 @@ def __delitem__(self, key: str) -> None:
         # the dictionary is not modified
         self._maybe_unset_graph(value)
         super().__delitem__(key)
+
+    def add(self, value: _core.Value) -> None:
+        """Add an initializer to the graph."""
+        self[value.name] = value  # type: ignore[arg-type]
```
justinchuby, line 272:
```suggestion
        self[value.name] = value  # type: ignore[index]
```

src/onnx_ir/_graph_containers.py
```diff
@@ -266,3 +270,88 @@ def __delitem__(self, key: str) -> None:
...
+            raise TypeError(f"Key must be a string, not {type(key)}")
+        if not isinstance(value, _core.Attr):
+            raise TypeError(f"Value must be an Attr, not {type(value)}")
+        super().__setitem__(key, value)
+
+    def add(self, value: _core.Attr) -> None:
+        """Add an attribute to the node."""
+        self[value.name] = value
+
+    def get_int(self, key: str, default: T = None) -> int | T:
```
justinchuby, line 297: @bmehta001 hopefully this makes getting attribute values easier?

src/onnx_ir/_graph_containers.py
```diff
@@ -266,3 +270,88 @@ def __delitem__(self, key: str) -> None:
...
+
+    def __setitem__(self, key: str, value: _core.Attr) -> None:
+        """Set an attribute for the node."""
+        if type(key) is not str:
+            raise TypeError(f"Key must be a string, not {type(key)}")
+        if not isinstance(value, _core.Attr):
+            raise TypeError(f"Value must be an Attr, not {type(value)}")
+        super().__setitem__(key, value)
+
+    def add(self, value: _core.Attr) -> None:
```
justinchuby, line 293: Check type

src/onnx_ir/passes/common/inliner.py
```diff
@@ -96,6 +96,7 @@ def clone_attr(self, key: str, attr: ir.Attr) -> ir.Attr | None:
             return attr
         assert attr.is_ref()
         ref_attr_name = attr.ref_attr_name
+        assert ref_attr_name is not None, "Reference attribute must have a name"
```
titaiwangms, line 99: We still have RefAttr?

src/onnx_ir/_graph_containers.py
```diff
@@ -250,6 +252,8 @@ def __setitem__(self, key: str, value: _core.Value) -> None:
             )
         if not isinstance(key, str):
             raise TypeError(f"Key must be a string, not {type(key)}")
+        if not value.name:
+            raise ValueError(f"Initializer must have a name: {value!r}")
```
titaiwangms, line 256: Should we set the key to the value name instead of crashing? I was a bit confused when I realized value.name and key could be set separately but not allowed to be different. It becomes a burden for users to be cautious and manually maintain.

src/onnx_ir/passes/common/inliner.py
justinchuby: Right, we combined the two classes so even though we don’t have the RefAttr class anymore, there is still a way to represent ref attr. And in this case we need to handle them.

src/onnx_ir/_core.py
```diff
@@ -2263,15 +2261,11 @@ def register_initializer(self, value: Value) -> None:
                     " it is not the same object: existing={self._initializers[value.name]!r},"
                     f" new={value!r}"
                 )
-        if value.producer() is not None:
```
justinchuby, line 2266: This is checked by the GraphInitializers class

titaiwangms (approved):

PR #35, opened 2025-05-30
Remove RefAttr from docs because it is not a class

justinchuby:


titaiwangms:
Probably go through all the assertion?

titaiwangms (approved):

PR #45, opened 2025-06-01
Convenience function get_const_tensor

justinchuby:
Get the constant tensor from a value, if it exists.

A constant tensor can be obtained if the value has a ``const_value`` set
(as in the case of an initializer) or if the value is produced by a
Constant node.

This function will not alter the ``const_value`` of the value, but
it will propagate the shape and type of the constant tensor to the value
if `propagate_shape_type` is set to True.


justinchuby:
@codecov-ai-reviewer test

src/onnx_ir/_convenience/__init__.py
```diff
@@ -389,3 +391,71 @@ def replace_nodes_and_values(
...
+            )
+        elif attr_name == "value":
+            tensor = attr_value.as_tensor()
+        else:
+            return None
+    tensor.name = value.name
+    if tensor is not None and propagate_shape_type:
+        # Propagate the shape and type of the tensor to the value
+        value.shape = tensor.shape  # type: ignore[assignment]
+        value.type = _core.TensorType(tensor.dtype)
```
titaiwangms, line 460: Is this the only way to propagate the shape and type to the value? I am wondering how value and constant have different shape and type at the first place?

src/onnx_ir/_convenience/__init__.py
```diff
@@ -389,3 +391,71 @@ def replace_nodes_and_values(
...
+
+    Args:
+        value: The value to get the constant tensor from.
+        propagate_shape_type: If True, the shape and type of the value will be
+            propagated to the Value.
+
+    Returns:
+        The constant tensor if it exists, otherwise None.
+    """
+    tensor = None
```
titaiwangms, line 417: It would be good to annotate. I found it can be assigned name.

titaiwangms (approved):
Unblocking nits and questions

src/onnx_ir/_convenience/__init__.py
```diff
@@ -389,3 +391,71 @@ def replace_nodes_and_values(
...
+    if value.const_value is not None:
+        tensor = value.const_value
+    else:
+        node = value.producer()
+        if node is None:
+            return None
```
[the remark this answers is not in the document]
justinchuby: This would an input. Updated

src/onnx_ir/_convenience/__init__.py
justinchuby: I think it is possible that the value may not have a shape and type set. Maybe we can check for that

src/onnx_ir/_convenience/__init__.py
justinchuby: Updated to assign name only for the new tensor (from constant)

src/onnx_ir/_convenience/__init__.py
```diff
@@ -389,3 +391,71 @@ def replace_nodes_and_values(
...
+    (as in the case of an initializer) or if the value is produced by a
+    Constant node.
+
+    This function will not alter the ``const_value`` of the value, but
+    it will propagate the shape and type of the constant tensor to the value
+    if `propagate_shape_type` is set to True.
+
+    Args:
+        value: The value to get the constant tensor from.
+        propagate_shape_type: If True, the shape and type of the value will be
```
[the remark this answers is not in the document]
justinchuby: I think this is the same situation. The basic propagation utility in onnxscript basically does this, which I promoted as a general utility here

PR #49, opened 2025-06-02
[Easy] Update name as ir-py instead of onnx-ir

jcwchen:
onnx-ir looks great! This PR is to solve some easy issues when browsing the repo.


justinchuby (approved):
Thanks!

justinchuby:
@jcwchen can you fix dco? Thanks

PR #51, opened 2025-06-02
Prevent values produced by other nodes to be added as graph inputs

justinchuby:


src/onnx_ir/_graph_containers.py
```diff
@@ -152,6 +152,8 @@ def _set_graph(self, value: _core.Value) -> None:
             raise ValueError(
                 f"Value '{value}' is already owned by a different graph. Please remove the value from the previous graph first"
             )
+        if value.producer() is not None:
+            raise ValueError(f"Value '{value}' is produced by a node and cannot be an input to the graph)
```
justinchuby, line 156:
```suggestion
            raise ValueError(f"Value '{value}' is produced by a node and cannot be an input to the graph")
```

PR #63, opened 2025-06-03
Introduce node convenience functions

Johansmm:
Convenience function to insert a set of nodes in an ir.Value(s).


src/onnx_ir/_convenience/__init__.py
```diff
@@ -367,11 +380,113 @@ def replace_nodes_and_values(
...
+    all_outputs = dict.fromkeys(sum((node.outputs for node in nodes), ()))  # type: ignore[type-var]
+    # A value is considered as input if it is not any output.
+    inputs = tuple(val for val in all_inputs if val not in all_outputs)
+    # A value is considered as output if it is not any input.
+    outputs = tuple(val for val in all_outputs if val not in all_inputs)
+    return inputs, outputs
+
+
+def insert_nodes_in_value(
+    values: _core.Value | Sequence[_core.Value], new_nodes: Sequence[_core.Node]
```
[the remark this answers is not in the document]
Johansmm: I am thinking in some case that is needed to insert new nodes without remove previous ones. Following: https://github.com/microsoft/onnxscript/issues/2064 discussion, this is not possible to do this with onnxscript.rewriter, since rewrite_pattern have to redefine target_pattern.

Moreover I believe with this helper it is easer to include new nodes with something like

```python 
for node in ir_model.graph:
    if node.op_type == 'expected_op_type':
        insert_node(node.outputs[0], new_node)
```

justinchuby:
I wonder if something like this would make sense:

```py
def insert_nodes_at(values, nodes, incoming_values, outgoing_values)
```

or if the new nodes are connected linearly

```
def insert_nodes_at(values, nodes)
```

(we can constraint the inputs can only be the first node, and the outputs can only be produced by the last node so that the intention is clear)

src/onnx_ir/_convenience/__init__.py
```diff
@@ -367,11 +380,113 @@ def replace_nodes_and_values(
...
+    replace_all_uses_with(inputs, values)
+
+    # Update graph if there is one:
+    if (graph := values[-1].graph) is not None:
+        # Update graph/function outputs if the node generates output
+        _update_graph_or_function_outputs(graph, values, outputs)
+
+        # Insert new nodes if there is a graph
+        graph.extend(new_nodes)
+        graph.sort()
```
justinchuby, line 492: sorting after insertion is not ideal. Is there a way to efficiently find the insertion point?

Johansmm:
> Thanks for the reminder - I will need to think about this more carefully once I am done with the thing I am currently working on. I will get back to you hopefully by the end of this week? Your comment makes sense. Are there special consideration needed to ensure node order? (topologically sorted and deterministic). Also we need to ensure there is no loops and the values fully encloses the subgraph?

Sorry @justinchuby I forget to answer you:
1. I do not think so. To deduce the nodes through `incoming_values` and `outgoing_values` It should be enough to go from `incoming_values` in the graph to `outgoing_values`, right? in this process the nodes would be added in a topological way.

In the same way when there are multiple incoming/outgoing, I think traversing the graph to deduce the nodes would force to sort them indirectly.
2. For the loops yes, we should verify there are no enclosures. I believe this should not be a problem.

src/onnx_ir/_convenience/__init__.py
Johansmm: Remove it

Johansmm:
Hi @justinchuby, I'm back with this idea (sorry for the delay). I'd like to know how to rethink the tool to make this possible.

I just did a push force to rebase in last main and removing `graph.sort()`.

PR #75, opened 2025-06-13
Improve docstrings

justinchuby:


justinchuby:
Add a note on ir.Node to refer to ir.node

justinchuby:
@titaiwangms thanks!

PR #80, opened 2025-06-15
Add tests for from_onnx_text method

justinchuby:
Add the missing tests.

### DIFF
--- a/src/onnx_ir/__init__.py
+++ b/src/onnx_ir/__init__.py
@@ -83,6 +83,7 @@ __all__ = [
     "save",
     # Flags
     "DEBUG",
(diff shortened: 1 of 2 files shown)


justinchuby:
There is a bit of ambiguity with the naming of "initializers". Here these are more like "tensors that will initialize values" with matching names that we can use to find the set the const_value of corresponding values. In graph the initializers provided in `__init__` is actually "initialized_values". I wasn't clear about this when designing the graph object initially. But then if we name the attributes in graph "initialized_values", we will also introduce a new concept that people have not seen, so the name "initializers" may be acceptable.

titaiwangms (approved):

Issue #57, opened 2025-06-02
Integrate symbolic shape inference


justinchuby:
Port and integrate symbolic shape inference from ONNX Runtime as an analysis pass / feature

inisis:
@justinchuby Can I have a short online meeting with you.

inisis:
No, can I connect you on linkedin, and arrange the meeting on google meet.

inisis:
I reclaimed my slack accout, let's connect.

justinchuby:
Done

PR #82, opened 2025-06-15
Support symbolic shape inference

inisis:
Closes https://github.com/onnx/ir-py/issues/57

1. Migrate symbolic shape inference code from onnxruntime;
2. Prefer symbolic shape inference over onnx built-in shape inference.

### DIFF
--- a/src/onnx_ir/passes/common/shape_inference.py
+++ b/src/onnx_ir/passes/common/shape_inference.py
@@ -15,6 +15,7 @@ import onnx
 
 import onnx_ir as ir
 from onnx_ir.passes.common import _c_api_utils
+from .symbolic_shape_infer import SymbolicShapeInference
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


justinchuby:
Thank you! A few suggestions:

1. We would like the `src/onnx_ir/passes/common/symbolic_shape_infer.py` module to be refactored and use only `onnx_ir` methods for graph manipulation. This would be the most important can time consuming part. But it's a good opportunity to do necessary refactoring of the logic. You may start with removing imports from onnx and see how we can do things with onnx_ir.
    a. We can consider using the onnx test data for validating the new logic. It can be done in a separate PR.
3. The inference methods need to be robust and tested. Is there a way to explicitly specify in each method the supported opset versions? It would be preferable if relevant information are written in a declarative manner.
4. Because the module is big, it may be helpful if you create a root directory `_symbolic_shape_inference` to host the code and break the logic into smaller files. When the logic matures we can expose the `ymbolic_shape_inference` module publically.
	a. It's preferable if you can modularize the original big chunk of logic so that they are individually testable and maintainable.
5. For the pass we would like to keep the old behavior. I suggest creating a new in-place `SymbolicShapeInferencePass` that will call the new logic, so users have a choice.

inisis:
Hi @justinchuby, I may not have enough bandwidth to do the refactoring, It's really time consuming😔.

justinchuby:
Thanks for letting me know! I am creating https://github.com/onnx/ir-py/pull/117 to continue.

PR #86, opened 2025-06-17
Port onnx float16 from onnxconverter-common

bjeffrey92:
A few weeks ago I created [an issue](https://github.com/onnx/onnx/discussions/7022) on the `onnx` library repo enquiring about the state of the `onnxconverter-common` library, which appears to no longer be actively maintained. @justinchuby responded to me and suggested that the module of `onnxconverter-common` which I require, for converting a single precision onnx model to mixed precision, could be ported into this library.

Here, I have ported the relevant functions from the `float16` module from `onnxconverter-common` (https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/float16.py) to this repo at the location suggested by @justinchuby. I also copied the relevant tests across.

Note, that this is not originally my code, hence I have retained the attribution and the original MIT license information with the source code. The modifications I made were just to:
- strip out some functions that are not needed for my use case
- add type hinting throughout
- apply formatting and linting rules specified in the ir-py repo
I have not made any material change to the logic of any of the functions in the original module in `onnxconverter-common`.


justinchuby:
You may also refer to the rest of the files in the passes directory for examples

bjeffrey92:
@justinchuby thanks for looking at this so quickly, but can you be a bit more specific please? I don't see any functions in the other modules of the passes directory which look like they replicate the `onnx.helper` and `onnx.numpy_helper` functions that are used here (`onnx.helper.make_node`, `onnx.helper.make_tensor_value_info`, `onnx.numpy_helper.to_array`)

bjeffrey92:
@justinchuby and when you say that you want me to remove the `onnx_proto` import, do you mean just import the protos from the top level of the `onnx` library, eg `onnx.ModelProto` rather `onnx.onnx_pb.ModelProto`? 

Like how it is referenced here: https://github.com/onnx/ir-py/blob/main/src/onnx_ir/passes/common/shape_inference.py#L22

justinchuby:
Specifically you may use:

onnx.helper.make_node: ir.node()
onnx.helper.make_tensor_value_info: value.shape=..., value.dtype=...
onnx.numpy_helper.to_array: ir.tensor()

PR #87, opened 2025-06-17
Ban onnx imports

justinchuby:
In all places except for tests, serde, and selected passes and helpers.

Also cleaned up src/onnx_ir/_version_utils.py


titaiwangms (approved):
What is the motivation?

PR #92, opened 2025-06-17
[Pass] Support CSE constant nodes

titaiwangms:


titaiwangms:
@justinchuby Not sure why I checked the sign off, but DCO still fails...

src/onnx_ir/passes/common/common_subexpression_elimination.py
```diff
@@ -17,93 +17,117 @@
...
+                ):
+                    # For INT, FLOAT and STRING attributes, we convert them to tuples
+                    # to ensure they are hashable.
+                    value = tuple(value)
+                elif v.type is ir.AttributeType.TENSOR:
+                    np_value = value.numpy()
+                    if np_value.nbytes > self.size_limit:
+                        # If the tensor is larger than the size limit, we skip it.
+                        large_tensor = True
+                        break
```
justinchuby, line 88:
```suggestion
                    if value.size > self.size_limit:
                        # If the tensor is larger than the size limit, we skip it.
                        large_tensor = True
                        break
                    np_value = value.numpy()

```

1. Compare the size (number of elements) instead of the bytes according to the intention
2. We should compare before calling numpy(). There is a cost to .numpy() but none for getting `value.size`

src/onnx_ir/passes/common/common_subexpression_elimination_test.py
```diff
@@ -321,3 +331,43 @@ def test_model(x: FLOAT[2, 2]) -> FLOAT[2, 2]:
...
+        """
+
+        @script()
+        def test_model(x: FLOAT[2, 2]) -> FLOAT[2, 2]:
+            a = op.Add(x, op.Constant(value=np.array([1.0, 2.0], dtype=np.float32)))
+            b = op.Add(x, op.Constant(value=np.array([1.0, 2.0], dtype=np.float32)))
+            return a + b
+
+        model_proto = test_model.to_model_proto()
+        model = ir.serde.deserialize_model(model_proto)
```
justinchuby, line 351: Comment on why round tripping is needed?

justinchuby:
Very nice, thanks!

justinchuby:
@titaiwangms maybe ensure that the email you signoff with matches the github account you use?

src/onnx_ir/passes/common/common_subexpression_elimination.py
```diff
@@ -17,93 +17,117 @@
...
+            # Skip large tensors to avoid cse weights and bias.
+            large_tensor: bool = False
+            # Use equality to check if the node is a common subexpression.
+            attributes = {}
+            for k, v in node.attributes.items():
+                # TODO(exporter team): CSE subgraphs.
+                # NOTE: control flow ops like Loop and If won't be CSEd
+                # because attribute: graph won't match.
+                if v.type in (ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS):
+                    control_flow_op = True
```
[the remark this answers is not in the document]
justinchuby: > There are ops that has graph but not control flow

Maybe? But either way is ok with me. Maybe keep what you have right now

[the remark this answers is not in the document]
titaiwangms: https://github.com/microsoft/onnxscript/pull/2399

src/onnx_ir/passes/common/common_subexpression_elimination.py
```diff
@@ -17,93 +17,118 @@
 
 
 class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
-    """Eliminate common subexpression in ONNX graphs."""
+    """Eliminate common subexpression in ONNX graphs.
+
+    Attributes:
+        size_limit: The minimum size of the tensor to be csed. If the tensor contains
+            number of elements larger than size_limit, it will not be csed. Default is 1024.
```
justinchuby, line 24:
```suggestion
            number of elements larger than size_limit, it will not be cse'd. Default is 1024.
```

nit

src/onnx_ir/passes/common/common_subexpression_elimination.py
```diff
@@ -17,93 +17,118 @@
...
-        control_flow_op: bool = False
-        # Use equality to check if the node is a common subexpression.
-        attributes = {}
-        for k, v in node.attributes.items():
-            # TODO(exporter team): CSE subgraphs.
-            # NOTE: control flow ops like Loop and If won't be CSEd
-            # because attribute: graph won't match.
-            if v.type in (ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS):
-                control_flow_op = True
+    def _eliminate_common_subexpression(self, graph: ir.Graph, modified: bool) -> bool:
```
justinchuby, line 45: Does this need to be a method? I was wondering what has changed. Can size_limit be passed in as well?

justinchuby, line 45: I also wonder why modified is needed as a parameter?

justinchuby (approved):

src/onnx_ir/passes/common/common_subexpression_elimination.py
titaiwangms: Should be doable. Is there a reason we prefer function to method?

justinchuby:
@gramalingam for another review

src/onnx_ir/passes/common/common_subexpression_elimination_test.py
```diff
@@ -23,6 +29,8 @@ def check_graph(self, model: ir.Model, inputs: list[ir.Value], delta_nodes: list
             delta_nodes: The expected change in the number of nodes in the model.
                          The length of this list should match the number of graphs
                          in the model. (to support subgraphs in the future)
+            size_limit: The minimum size of the tensor to be csed. If the tensor contains
```
gramalingam, line 32: Maximum

gramalingam (approved):
LGTM, modulo couple of minor nits

src/onnx_ir/passes/common/common_subexpression_elimination.py
```diff
@@ -17,93 +17,117 @@
 
 
 class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
-    """Eliminate common subexpression in ONNX graphs."""
+    """Eliminate common subexpression in ONNX graphs.
+
+    Attributes:
+        size_limit: The minimum size of the tensor to be csed. If the tensor contains
+            number of elements larger than size_limit, it will not be cse'd. Default is 1024.
```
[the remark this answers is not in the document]
titaiwangms: Done

PR #94, opened 2025-06-18
Create adapters for torch

justinchuby:
Fix https://github.com/onnx/ir-py/issues/93

### DIFF
--- a/docs/api/ir_tensor_adapters.md
+++ b/docs/api/ir_tensor_adapters.md
@@ -4,7 +4,7 @@
 .. automodule:: onnx_ir.tensor_adapters
 ```
 
-## Adaptors for PyTorch
+## Adapters for PyTorch
 
 ```{eval-rst}
 .. autosummary::
--- a/docs/api/ir_tensor_adapters.md
+++ b/docs/api/ir_tensor_adapters.md
@@ -1,0 +1,21 @@
+# onnx_ir.tensor_adapters
+
+```{eval-rst}
+.. automodule:: onnx_ir.tensor_adapters
+```
+
+## Adaptors for PyTorch
+
+```{eval-rst}
+.. autosummary::
+    :toctree: generated
+    :template: classtemplate.rst
+    :nosignatures:
+
+    onnx_ir.tensor_adapters.TorchTensor
+```
+
+```{eval-rst}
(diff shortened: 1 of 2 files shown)


docs/api/ir_tensor_adapters.md
```diff
@@ -0,0 +1,21 @@
+# onnx_ir.tensor_adapters
+
+```{eval-rst}
+.. automodule:: onnx_ir.tensor_adapters
+```
+
+## Adaptors for PyTorch
```
justinchuby, line 7:
```suggestion
## Adapters for PyTorch
```

src/onnx_ir/tensor_adapters.py
```diff
@@ -56,7 +56,7 @@ def from_torch_dtype(dtype: torch.dtype) -> ir.DataType:
     if _TORCH_DTYPE_TO_ONNX is None:
         import torch
 
-        _TORCH_DTYPE_TO_ONNX: dict[torch.dtype, ir.DataType] = {
```
titaiwangms, line 54: The annotation does not work?

src/onnx_ir/tensor_adapters.py
```diff
@@ -72,8 +77,58 @@ def __init__(
...
+            ir.DataType.FLOAT8E5M2: torch.float8_e5m2,
+            ir.DataType.FLOAT8E5M2FNUZ: torch.float8_e5m2fnuz,
+            ir.DataType.INT16: torch.int16,
+            ir.DataType.INT32: torch.int32,
+            ir.DataType.INT64: torch.int64,
+            ir.DataType.INT8: torch.int8,
+            ir.DataType.UINT8: torch.uint8,
+            ir.DataType.UINT16: torch.uint16,
+            ir.DataType.UINT32: torch.uint32,
+            ir.DataType.UINT64: torch.uint64,
```
titaiwangms, line 114: Is this just the flip of the _TORCH_DTYPE_TO_ONNX? If so,  maybe we can generate this one from it?

justinchuby: Kinda hard because we want to import torch locally so

src/onnx_ir/tensor_adapters.py
justinchuby: Also they are annotated above

PR #97, opened 2025-06-19
Set up copilot

justinchuby:


titaiwangms (commented):
Can you put the reference on description. I don't see how this enables Copilot.

PR #105, opened 2025-06-23
Document changes between versions in docstrings

justinchuby:
With sphinx versionchanged and versionadded directives


titaiwangms (approved):
From now on, we need to log whenever a new version is about to be released? Is there an instruction showing how should we maintain this?

titaiwangms:
> > From now on, we need to log whenever a new version is about to be released? Is there an instruction showing how should we maintain this?
> 
> I agree it would be a good practice to document important API changes. I can add a line in https://github.com/onnx/ir-py/wiki/Coding-style for now and move to CONTRIBUTIONS.md. Do you have any suggestions?

SG!

PR #114, opened 2025-06-27
Create py.typed to allow type checkers

justinchuby:
In https://github.com/pytorch/pytorch/pull/157017#discussion_r2172124344, I realized we need this marker file to indicate that type checkers can check types on this package. It will be automatically include by setuptools


justinchuby:
@titaiwangms

titaiwangms (approved):

justinchuby:
Sorry I put it in the wrong directory

PR #117, opened 2025-06-30
[WIP] Create symbolic type/shape inference logic

justinchuby:


justinchuby:
Replaced by https://github.com/onnx/ir-py/pull/327

PR #130, opened 2025-07-18
Fix string and undefined in tobytes()

titaiwangms:
STRING and UNDEFINED are not supported by itemsize.

NOTE: STRING shows up when array[None] appears (UNDEFINED is added only for consistency.)


src/onnx_ir/_core.py
```diff
@@ -512,7 +512,10 @@ def tobytes(self) -> bytes:
         }:
             # Pack the array into int4
             array = _type_casting.pack_4bitx2(array)
-        else:
+        elif self.dtype not in {
+            _enums.DataType.STRING,
+            _enums.DataType.UNDEFINED,
+        }:
```
justinchuby, line 518: I think tobytes is not defined on these two dtypes. Curious when we would hit this?

justinchuby:
Is tobytes() well defined for STRING? I wonder why we would see array[None]? Is it valid?

justinchuby:
Closed because strings and undefined values should not have bytes representation

PR #133, opened 2025-07-24
Create DeduplicateBigInitializersPass

justinchuby:
I implemented hashing on DeduplicateBigInitializersPass to handle bigger tensors, inspired by https://github.com/huggingface/optimum-onnx/blob/90ef2106a43393a88745e3259535ad86c24f3793/optimum/onnx/transformations_utils.py#L58

### DIFF
--- a/src/onnx_ir/passes/common/initializer_deduplication.py
+++ b/src/onnx_ir/passes/common/initializer_deduplication.py
@@ -4,14 +4,17 @@
 
 from __future__ import annotations
 
-import hashlib
-
 __all__ = [
     "DeduplicateInitializersPass",
 ]
 
 
+import hashlib
+import logging
+
 import onnx_ir as ir
+
+logger = logging.getLogger(__name__)
 
 
 class DeduplicateInitializersPass(ir.passes.InPlacePass):
@@ -62,10 +65,14 @@ class DeduplicateInitializersPass(ir.passes.InPlacePass):
             key = (const_val.dtype, tensor_dims, tensor_digest)
 
             if key in initializers:
-                assert initializers[key].const_value.tobytes() == const_val.tobytes(), (
-                    "Initializer deduplication failed: "
-                    f"hashes match but values differ with values {initializers[key]} and {initializer}"
-                )
+                if initializers[key].const_value.tobytes() != const_val.tobytes():
+                    logger.warning(
+                        "Initializer deduplication failed: "
+                        "hashes match but values differ with values %s and %s",
+                        initializers[key],
+                        initializer
+                    )
+                    continue
                 modified = True
                 ir.convenience.replace_all_uses_with(initializer, initializers[key])  # type: ignore[index]
                 assert initializer.name is not None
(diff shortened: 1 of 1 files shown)


src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -44,8 +50,29 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
             if const_val.size > self.size_limit:
                 continue
 
-            key = (const_val.dtype, tuple(const_val.shape), const_val.tobytes())
+            if const_val.dtype == ir.DataType.STRING:
+                # Skip string initializers as they don't have a bytes representation
+                continue
+
+            # Hash tensor data to avoid storing large amounts of data in memory
```
gramalingam, line 57: I am not sure I understand the complexity/perf implications or claim: computing a hash-value requires a complete linear pass over the tensor-data, doesn't it? Why would we do that eagerly? Unless users explicitly ask for it?

Arguably, a comparison between different tensors will likely fail very early (and could be more efficient than computing a hash-value eagerly).

justinchuby: Hmm that's true

justinchuby: I created a separate DeduplicateBigInitializersPass. Hopefully this allows users to explicitly call it when needed.

src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -54,3 +61,67 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
                 initializers[key] = initializer  # type: ignore[index]
 
         return ir.passes.PassResult(model=model, modified=modified)
+
+
+class DeduplicateBigInitializersPass(ir.passes.InPlacePass):
+    """Remove big, duplicated initializer tensors from the graph.
```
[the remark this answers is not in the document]
justinchuby: Updated to DeduplicateHashedInitializersPass. Would that be more clear?

src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -54,3 +61,68 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+
+
+class DeduplicateHashedInitializersPass(ir.passes.InPlacePass):
+    """Remove duplicated initializer tensors (using a hashed method) from the graph.
+
+    This pass detects initializers with identical shape, dtype, and hashed content,
+    and replaces all duplicate references with a canonical one.
+
+    This pass should have a lower peak memory usage than :class:`DeduplicateInitializersPass`
+    as it does not store the full tensor data in memory, but instead uses a hash of the tensor data.
```
gramalingam, line 73: Are the initializers loaded on-demand? Or, is the model expected to be loaded with all external tensors initially (which would mean no improvements to peak-memory usage.)

In general, does the IR expose a way for users to request loading/freeing an external tensor? That would enable sophisticated algorithms that want to handle them carefully to control peak memory.

PR #156, opened 2025-08-19
[pass] Deduplicate initializers in subgraphs and fix bug in `subgraph()`

justinchuby:
- Deduplicate initializers in subgraphs in the DeduplicateInitializersPass
- Fix the pass to correctly skip graph inputs and outputs and STRING tensors.
- Fix a bug in `graph.subgraphs()` where graphs with no nodes are omitted in the iterator.

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -2571,6 +2571,10 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
         def enter_subgraph(graph) -> None:
             if graph is self:
                 return
+            if not isinstance(graph, Graph):
+                raise TypeError(
+                    f"Expected a Graph, got {type(graph)}. The model may be invalid"
+                )
             if graph not in seen_graphs:
                 seen_graphs[graph] = None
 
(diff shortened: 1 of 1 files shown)


src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -9,48 +9,63 @@
...
+                    # Skip graph inputs and outputs
+                    logger.warning(
+                        "Skipped deduplication of initializer '%s' as it is a graph input or output",
+                        initializer.name,
+                    )
+                    continue
+
+                const_val = initializer.const_value
+                if const_val is None:
+                    # Skip if initializer has no constant value
```
titaiwangms, line 50: Should we log this?

titaiwangms (commented):
We need a test.

justinchuby:
Which logic block do you want to see more tests for?

gramalingam (approved):

src/onnx_ir/_core.py
```diff
@@ -2564,14 +2564,18 @@ def subgraphs(self) -> Iterator[Graph]:
 
         .. versionadded:: 0.1.2
         """
-        seen_graphs: set[Graph] = set()
-        for node in onnx_ir.traversal.RecursiveGraphIterator(self):
-            graph = node.graph
+        seen_graphs: dict[Graph, None] = {}
```
justinchuby, line 2567:
```suggestion
        # Use a dict to preserve order
        seen_graphs: dict[Graph, None] = {}
```

PR #169, opened 2025-08-27
Support string tensors in DeduplicateInitializersPass and call_onnx_api

iksnagreb:
Getting the number of bytes in a string tensor needs special treatment as the STRING data type does not define a bitwidth but needs to be computed from flattening the strings into a sequence of bytes. See call_onnx_api querying .nbytes to temporarily remove large initializers.

String initializers need to be converted to bytes via the .string_data method instead of the .tobytes method in DeduplicateInitializersPass.

Also adds mapping of `np.bytes_` to `STRING` in `DataType.from_numpy`, which was probably just overlooked before, as object and `np.str_` (unicode strings) are already handled. This is related to https://github.com/microsoft/onnxscript/pull/2514

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -837,6 +837,16 @@ class StringTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=to
         return self._shape
 
     @property
+    def size(self) -> int:
+        """The number of elements in the tensor."""
+        return sum((len(string) for string in self.string_data()))
+
+    @property
+    def nbytes(self) -> int:
+        """The number of bytes in the tensor."""
+        return self.size
+
+    @property
     def raw(self) -> Sequence[bytes] | npt.NDArray[np.bytes_]:
         """Backing data of the tensor. Immutable."""
         return self._raw  # type: ignore[return-value]
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -837,14 +837,9 @@ class StringTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=to
         return self._shape
 
     @property
-    def size(self) -> int:
-        """The number of elements in the tensor."""
-        return sum(len(string) for string in self.string_data())
-
-    @property
     def nbytes(self) -> int:
         """The number of bytes in the tensor."""
-        return self.size
+        return sum(len(string) for string in self.string_data())
 
     @property
     def raw(self) -> Sequence[bytes] | npt.NDArray[np.bytes_]:
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -839,7 +839,7 @@ class StringTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=to
     @property
     def size(self) -> int:
         """The number of elements in the tensor."""
-        return sum((len(string) for string in self.string_data()))
+        return sum(len(string) for string in self.string_data())
 
     @property
     def nbytes(self) -> int:
--- a/src/onnx_ir/_enums.py
+++ b/src/onnx_ir/_enums.py
@@ -77,7 +77,7 @@ class DataType(enum.IntEnum):
         if dtype in _NP_TYPE_TO_DATA_TYPE:
             return cls(_NP_TYPE_TO_DATA_TYPE[dtype])
 
-        if np.issubdtype(dtype, np.str_):
+        if np.issubdtype(dtype, np.str_) or np.issubdtype(dtype, np.bytes_):
             return DataType.STRING
 
         # Special cases for handling custom dtypes defined in ONNX (as of onnx 1.18)
--- a/src/onnx_ir/_enums.py
+++ b/src/onnx_ir/_enums.py
@@ -215,6 +215,10 @@ class DataType(enum.IntEnum):
             DataType.FLOAT8E8M0,
         }
 
+    def is_string(self) -> bool:
+        """Returns True if the data type is a string type."""
+        return self == DataType.STRING
+
     def __repr__(self) -> str:
         return self.name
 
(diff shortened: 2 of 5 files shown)


src/onnx_ir/_core.py
```diff
@@ -836,6 +836,16 @@ def shape(self) -> Shape:
         """The shape of the tensor. Immutable."""
         return self._shape
 
+    @property
+    def size(self) -> int:
+        """The number of elements in the tensor."""
+        return sum(len(string) for string in self.string_data())
```
justinchuby, line 842: I think this should be len(self.string_data()) right? Because it's counting elements

src/onnx_ir/_core.py
```diff
@@ -836,6 +836,16 @@ def shape(self) -> Shape:
...
+
+    @property
+    def nbytes(self) -> int:
+        """The number of bytes in the tensor."""
+        return self.size
```
justinchuby, line 847: It would be desirable if this is O(1). Is there a good way to do it?

iksnagreb: Hm, not sure, if I understand correctly, the .string_data() gives a list of strings (actually a `Sequence[bytes]`), so you have to ask each string for its length? The problem seems to be that _raw can be either a numpy array of bytes (these should know the total size) or the `Sequence[bytes]` (need to sum over length).

justinchuby:
Could you add a test in https://github.com/onnx/ir-py/blob/main/src/onnx_ir/passes/common/initializer_deduplication_test.py?

iksnagreb:
I wonder if there is a particular reason to not implement `.tobytes` for `StringTensor`? Otherwise the changes to the two deduplicate initializer passes could be unified by simply implementing the `.tobytes` method without weird workarounds checking for the data type just to do basically `.string_data().tobytes()` instead?

justinchuby:
The onnx proto def does not support string as bytes value. So I didn’t see a need for the method at the time. Maybe we can revisit this?

src/onnx_ir/passes/common/initializer_deduplication_test.py
```diff
@@ -60,6 +73,19 @@ def test_initializers_with_different_shapes_not_deduplicated(self):
         new_model = self.apply_pass(model)
         self.assertEqual(len(new_model.graph.initializers), 2)
 
+    def test_string_initializers_with_different_shapes_not_deduplicated(self):
+        model = ir.from_onnx_text(
+            """
+            <ir_version: 10, opset_import: ["" : 17]>
+            agraph () => ()
+            <string[2] s1 = {"A", "B"}, string[1,2] s2 = {"A", "B"}> {
```
justinchuby, line 81: Could you also test the same shape but "ab", "c" vs "a", "bc" case? Thanks

justinchuby:
```
>>> np.array((b"a", b"b"))
array([b'a', b'b'], dtype='|S1')
>>> np.array((b"a", b"b")).tobytes()
b'ab'
```

This is what I got

src/onnx_ir/_core.py
```diff
@@ -836,6 +836,11 @@ def shape(self) -> Shape:
         """The shape of the tensor. Immutable."""
         return self._shape
 
+    @property
+    def nbytes(self) -> int:
+        """The number of bytes in the tensor."""
+        return sum(len(string) for string in self.string_data())
```
justinchuby, line 842: Would be great if you can test this too in `_core_test.py`

src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -42,17 +44,17 @@ def _should_skip_initializer(initializer: ir.Value, size_limit: int) -> bool:
...
-        )
-        return True
     return False
 
 
+def _tobytes(val: ir.TensorProtocol):
+    # StringTensor does not support tobytes. Use 'string_data'
+    # instead.
+    if val.dtype.is_string():
+        return np.array(val.string_data()).tobytes()
```
justinchuby, line 54: I think val.string_data() should yield a list of bytes right? So the array wouldn't have null terminators?

justinchuby: That said, if you want to decode the data and make it a string array in np which will add the null byte like your experiment suggested, I suppose that will work too?

iksnagreb: Hm, I don't thinks that's actually null terminators, but padding bytes so that each string occupies the same number of consecutive bytes in the flattened `.tobytes` representation. This way the flat bytes sequence together with the tensor shape should be enough to uniquely identify a string tensor and decoding does not require to find the per-element string length (whereas for arbitrary length, null-terminated strings this would be necessary).

Trying to summarize: the `.string_data` method corresponds to the referenced onnx spec above, the proposed `.tobytes` seems to be out of spec, so it should probably not be added to the `StringTensor` aiming to implement the onnx spec(?) and should not be used for serializing the tensor (as there could be differences in null-terminator and padding bytes and `.string_data`already implements that according to the spec), however, it is sufficient for the purpose of getting a key to uniquely identify tensors for deduplication. So maybe we could keep this local and give it another name to make the intended use clear?

src/onnx_ir/passes/common/initializer_deduplication.py
```diff
@@ -42,17 +44,17 @@ def _should_skip_initializer(initializer: ir.Value, size_limit: int) -> bool:
...
-            "Skipped deduplication of string initializer '%s' (unsupported yet)",
-            initializer.name,
-        )
-        return True
     return False
 
 
+def _tobytes(val: ir.TensorProtocol):
+    # StringTensor does not support tobytes. Use 'string_data'
+    # instead.
```
justinchuby, line 52: Could you add a comment on the padding part of it and say that the bytes representation is unambiguous but different from the raw string data, for future readers? Thanks a lot!

justinchuby:
Sorry for nitpicking but could you move the comment to inside the function definition? This way they can be seen more related

justinchuby:
Also commits need to be signed off

justinchuby (approved):

src/onnx_ir/_enums.py
```diff
@@ -215,6 +215,10 @@ def is_signed(self) -> bool:
             DataType.FLOAT8E8M0,
         }
 
+    def is_string(self) -> bool:
```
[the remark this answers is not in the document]
iksnagreb: Sorry, https://github.com/onnx/ir-py/pull/177

PR #171, opened 2025-08-28
Add more properties for numeric types to DataType

### LINKED ISSUE #109: Consider adding more properties to DataType
It would be convenient to augment the DataType enum to represent more properties of the data type (besides the bitwidth, name, and NumPy equivalent), such as signedness, whether it is an integer or float (the latter is already implemented), or the minimum/maximum supported value if relevant. These could then be considered, for example, in match conditions of ONNX Script pattern rewrite rules.

iksnagreb:
Fix #109


justinchuby:
Thanks for contributing these features!

PR #179, opened 2025-09-06
Create a convenience constructor for Value

justinchuby:
A convenience constructor for Value that supports an easier way of supplying arguments.

### DIFF
--- a/src/onnx_ir/_convenience/_constructors.py
+++ b/src/onnx_ir/_convenience/_constructors.py
@@ -215,3 +215,30 @@ def node(
         doc_string=doc_string,
         metadata_props=metadata_props,
     )
+
+
+def value(
+    name: str,
+    type: ir.Type,
+    doc_string: str | None = None,
+) -> ir.Value:
+    """Create a Value with the given name and type.
+
+    Example::
+
+        >>> import onnx_ir as ir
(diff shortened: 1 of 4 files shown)


src/onnx_ir/_convenience/_constructors.py
```diff
@@ -215,3 +215,66 @@ def node(
...
+
+        >>> import onnx_ir as ir
+        >>> t = ir.value("x", ir.DataType.FLOAT, ["N", 42, 3])
+        >>> t.name
+        'x'
+        >>> t.type
+        Tensor(FLOAT)
+        >>> t.shape
+        Shape([SymbolicDim(N), 42, 3])
+
```
justinchuby, line 240: Versonadded

src/onnx_ir/_convenience/_constructors.py
```diff
@@ -215,3 +215,66 @@ def node(
...
+def value(
+    name: str,
+    dtype: ir.DataType | None = None,
+    shape: ir.Shape | Sequence[int | str | None] | None = None,
+    *,
+    type: ir.TypeProtocol | None = None,
+    const_value: ir.TensorProtocol | None = None,
+) -> ir.Value:
+    """Create a Value with the given name and type.
```
justinchuby, line 228: Contrast with Value

justinchuby:
1. Call it `ir.val` to disambiguate from `Value`?
2. Deprecated Input

PR #197, opened 2025-09-26
Update the graph initializers dictionary when renaming Value

justinchuby:
This pull request enhances the handling of value renaming within the ONNX IR core: When a value (especially an initializer) is renamed, all relevant references—including the graph's initializer dictionary—are consistently updated. This reduces the risk of stale or inconsistent state when manipulating graph initializers.

Key changes:

**Core logic improvements:**

* The `Value` class now ensures that renaming a value also updates the corresponding entry in the graph's `initializers` dictionary, maintaining consistency between the value's name and its registration in the graph. An error is raised if an initializer's name is set to `None`.

* The `__init__` method docstring for `Value` has been updated to document this new behavior, clarifying that renaming a value also updates the graph initializer entry if applicable.

**Code simplification:**

* In `naming.py`, redundant logic for manually updating the `initializers` dictionary after a value rename has been removed, since this is now handled automatically by the `Value` class. This makes the code cleaner and less error-prone.

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -2031,6 +2031,14 @@ class Value(_protocols.ValueProtocol, _display.PrettyPrintable):
     ) -> None:
         """Initialize a value.
 
+        When assigning a name to the value, the name of the backing `const_value` (Tensor)
+        will also be updated. If the value is an initializer of a graph, the initializer
+        entry in the graph will also be updated.
+
+        .. versionchanged:: 0.1.10
+            Assigning a name to the value will also update the graph initializer entry
+            if the value is an initializer of a graph.
+
         Args:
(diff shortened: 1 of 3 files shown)


src/onnx_ir/_core.py
```diff
@@ -2025,6 +2025,14 @@ def __init__(
     ) -> None:
         """Initialize a value.
 
+        When assigning a name to the value, the name of the backing `const_value` (Tensor)
+        will also be updated. If the value is an initializer of a graph, the initializers
+        dictionary of graph will also be updated.
```
justinchuby, line 2030:
```suggestion
        dictionary of the graph will also be updated.
```

titaiwangms (approved):

PR #204, opened 2025-09-30
[pass] Preserve value metadata in LiftConstantsToInitializersPass

justinchuby:


src/onnx_ir/passes/common/constant_manipulation.py
```diff
@@ -69,7 +69,10 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
                 shape=tensor.shape,  # type: ignore[arg-type]
                 type=ir.TensorType(tensor.dtype),
                 const_value=tensor,
+                metadata_props=node.outputs[0].metadata_props.copy(),
             )
+            initializer.meta.update(node.outputs[0].meta)
```
titaiwangms, line 74: Should we add a simple comment like "preserve value meta for intermediate analysis". And for metadata_props like "preserve value info to its onnx model". I guess it wouldn't be needed if people all know what they are. But based on their names, I think people still need to spend extra time to refresh why we have two kinds of meta here. But this is not blocking this PR.

PR #235, opened 2025-10-28
Implement replace_all_uses_with on Value

justinchuby:
Given it is a fundamental operation in graph transformation, I added it as a method on `Value` so that it is more convenient and discoverable. The ir.convenience.replace_all_uses_with function is preserved to handle batch processing cases on multiple values.

Also added a `replace_graph_outputs` option to allow users to replace graph outputs as well. Users are still responsible to assign the output name to the replacement value if they want to preserve the signature.

### BC breaking

An error is raised when `replace_graph_outputs` is False && when the value to replace is a graph output.

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -2470,6 +2470,10 @@ class Value(_protocols.ValueProtocol, _display.PrettyPrintable):
             replacement: The value to replace all uses with.
             replace_graph_outputs: If True, graph outputs that reference this value
                 will also be updated to reference the replacement.
+
+        Raises:
+            ValueError: When ``replace_graph_outputs`` is False && when the value to
+                replace is a graph output.
         """
         # NOTE: Why we don't replace the value name when the value is an output:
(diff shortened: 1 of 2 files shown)


src/onnx_ir/_core.py
```diff
@@ -2448,6 +2448,42 @@ def is_initializer(self) -> bool:
...
+            replace_graph_outputs: If True, graph outputs that reference this value
+                will also be updated to reference the replacement.
+        """
+        # NOTE: Why we don't replace the value name when the value is an output:
+        # When the replacement value is already an output of the graph, renaming it
+        # to the name of this value will cause name conflicts. It is better to let
+        # the user handle the renaming explicitly and insert identity nodes if needed.
+        for user_node, index in self.uses():
+            user_node.replace_input_with(index, replacement)
+        if replace_graph_outputs and self.is_graph_output():
```
xadupre: Should we raise an exception when ``replace_graph_outputs is False and self.is_graph_output() is True``?

PR #238, opened 2025-10-31
Simplify rich display

justinchuby:
Since rich does not have the best color scheme for the onnx text, and may misinterpret things like `[...]`, I am removing the color output from the display() method

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -13,7 +13,6 @@
 from __future__ import annotations
 
 import abc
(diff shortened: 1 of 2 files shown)


justinchuby:
@gramalingam

gramalingam:
Actually, I liked the color display. It helps to have different colors. It seems like a simple fix might be to just change the str method for Shape to use "()" instead of "[]" or use any escape mechanism that rich provides. Seems a bit drastic to remove it completely.

PR #253, opened 2025-11-19
Fix serialization of attribute reference

gramalingam:
Fix bug in serialization logic of attribute reference ... it wasn't returning the serialized value.


justinchuby (approved):
Thanks!

PR #279, opened 2025-12-17
improve speed of op_identitifer

xadupre:
Node.op_identifier is used a lot when a model is optimized.

Before the change:

image

After the change:

image


justinchuby:
Is the picture showing time? Hit counts?

Issue #172, opened 2025-08-28
Introduce clone model


Johansmm:
I would like to know if it is possible to clone an `ir.Model` from API. I would like something like we can make in `onnx`:
```python
new_model = onnx.ModelProto()
new_model.CopyFrom(original_model)
```

Or simpler as:
```python
new_model_ir = original_model_ir.clone()
```

Today, (to make this) I have to go through `onnx`:
```python
def clone_ir_model(model):
    return ir.from_proto(ir.to_proto(model))
```

Which is not optimal

Johansmm:
Is this related to #58 ?

justinchuby:
Yes. We always wanted to create a clone method but didn't get to it yet. Contribution welcome if you are interested!

justinchuby:
Feel free to make a decision as the previous PR was copilot generated

PR #286, opened 2025-12-23
Implement graph cloning methods

justinchuby:
### Core IR API Enhancements

- Added `clone()` methods to the `Graph`, `Model`, and `Function` classes in `src/onnx_ir/_core.py`, enabling efficient deep copies of these objects using the new `Cloner` infrastructure.
- All mutable objects: nodes, values, graphs, shapes are duplicated. Tensor data and non-graph attributes (immutable) are shared.

### Cloning Infrastructure

- Introduced a new `Cloner` class by adopting and improving the cloner from inliner.
- Refactored the inliner pass in `src/onnx_ir/passes/common/inliner.py` to remove its custom `Cloner` implementation and use the new shared `Cloner` class, reducing code duplication and improving maintainability.

### Pass Infrastructure Improvements

- Added a `functionalize` utility in `src/onnx_ir/passes/_pass_infra.py` to easily convert in-place passes into functional passes by cloning the model before execution. 
- Updated pass module exports and imports to include the new `functionalize`.

Fix https://github.com/onnx/ir-py/issues/172

cc @iksnagreb @Johansmm for input, thanks.

### DIFF
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -2917,6 +2917,8 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
 
         Tensors in initializers and constant values will be shared.
 
+        .. versionadded:: 0.1.14
+
         Returns:
             A deep copy of this graph.
         """
@@ -3466,6 +3468,8 @@ Model(
 
         Tensors in initializers and constant values will be shared.
 
+        .. versionadded:: 0.1.14
+
         Returns:
             A deep copy of this model.
         """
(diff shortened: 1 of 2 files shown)


src/onnx_ir/_cloner.py
```diff
@@ -0,0 +1,187 @@
...
+            elif attr.type == _enums.AttributeType.GRAPHS:
+                graphs = [self.clone_graph(graph) for graph in attr.as_graphs()]
+                return _core.Attr(
+                    key, _enums.AttributeType.GRAPHS, graphs, doc_string=attr.doc_string
+                )
+            return attr
+
+        assert attr.is_ref()
+        if not self._resolve_ref_attrs:
+            return attr
```
gramalingam: So, an ir.Attr is an immutable object?

gramalingam: Allowing it to be shared by the clone, I mean?

gramalingam (approved):

PR #293, opened 2025-12-30
Allow `GraphView`s to be cloned

justinchuby:
This is useful for extracting a subgraph without breaking the original graph.


justinchuby:
Merged together with https://github.com/onnx/ir-py/pull/294

PR #294, opened 2025-12-30
Implement the graph extractor

justinchuby:
Extract subgraphs with `onnx_ir.convenience.extract()`. Migrated from https://github.com/microsoft/onnxscript/pull/2119.

Requires https://github.com/onnx/ir-py/pull/293

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,7 +77,7 @@ jobs:
         # To toggle linter comments in the files page, press `i` on the keyboard
         if: always()
         continue-on-error: true
-        uses: github/codeql-action/upload-sarif@v4.31.8
+        uses: github/codeql-action/upload-sarif@v4.31.9
         with:
           # Path to SARIF file relative to the root of the repository
           sarif_file: lintrunner.sarif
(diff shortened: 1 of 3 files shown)


src/onnx_ir/_convenience/_extractor.py
```diff
@@ -0,0 +1,191 @@
...
+            if val not in values:
+                raise ValueError(f"Value with name '{val}' not found in the graph.")
+
+    input_vals = [values[val] if isinstance(val, str) else val for val in inputs]
+    output_vals = [values[val] if isinstance(val, str) else val for val in outputs]
+    # Find the owning graph of the outputs to set as the parent graph
+    if not output_vals:
+        raise ValueError("At least one output must be provided to extract a subgraph.")
+    parent_graph = output_vals[0].graph
+    assert parent_graph is not None
```
gramalingam: Can't we use `graph` (from above) as the parent_graph? I think they should be the same?

justinchuby: graph may be a GraphView, which will not be the same graph as the owning graph of the values

justinchuby: Most of the places (all passes, rewriter etc.) we only handle graphs so it should be fine? Here I explicitly supported GraphView for more flexibility.

PR #313, opened 2026-01-14
Add allow_outer_scope_values option in Graph.clone()

justinchuby:
This pull request adds support for optionally preserving references to outer-scope values when cloning graphs in the ONNX IR, instead of always cloning them. This is particularly useful when cloning subgraphs that reference values defined in an outer graph. 

* Added a `allow_outer_scope_values` argument to the `Graph.clone()` method.
* Also preserve metadata.

### DIFF
--- a/src/onnx_ir/_cloner.py
+++ b/src/onnx_ir/_cloner.py
@@ -44,7 +44,7 @@ class Cloner:
         metadata_props: dict[str, str],
         post_process: Callable[[_core.Node], None] = lambda _: None,
         resolve_ref_attrs: bool = False,
-        no_outer_scope_values: bool = True,
+        allow_outer_scope_values: bool = False,
     ) -> None:
         """Initializes the cloner.
 
@@ -58,11 +58,11 @@ class Cloner:
                 processing on the cloned node.
             resolve_ref_attrs: Whether to resolve reference attributes using the attr_map.
                 Set to True when inlining functions.
-            no_outer_scope_values: When False, values that are from outer scopes
+            allow_outer_scope_values: When True, values that are from outer scopes
                 (not defined in this graph) will not be cloned. Instead, the cloned
                 graph will reference the same outer scope values. This is useful
                 when cloning subgraphs that reference values from the outer graph.
-                When True, values from outer scopes will cause an error if they
+                When False (default), values from outer scopes will cause an error if they
                 are referenced in the cloned graph.
         """
         self._value_map = value_map
@@ -70,7 +70,7 @@ class Cloner:
         self._metadata_props = metadata_props
         self._post_process = post_process
         self._resolve_ref_attrs = resolve_ref_attrs
-        self._no_outer_scope_values = no_outer_scope_values
+        self._allow_outer_scope_values = allow_outer_scope_values
 
     @_capture_error_context
     def _get_value(self, value: _core.Value) -> _core.Value:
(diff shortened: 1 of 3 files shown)


src/onnx_ir/_cloner.py
```diff
@@ -168,11 +199,15 @@ def clone_node(self, node: _core.Node) -> _core.Node:
     @_capture_error_context
     def clone_graph(self, graph: _core.Graph | _core.GraphView) -> _core.Graph:
         """Clones a graph with shared TensorProtocols."""
-        input_values = [self.clone_value(v) for v in graph.inputs]
+        if self._preserve_outer_scope_values:
+            # Collect local values only when we are preserving outer-scope values.
```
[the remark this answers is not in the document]
justinchuby: Excellent

src/onnx_ir/_cloner.py
```diff
@@ -58,29 +70,47 @@ def __init__(
...
+    def _clone_value(self, value: _core.Value) -> _core.Value:
+        value_map = self._value_map
+        if value in value_map:
+            known_value = value_map[value]
             assert known_value is not None, f"BUG: Value {value} mapped to None in value map"
             return known_value
-        # If the value is not in the value map, it must be a graph input.
-        # Note: value.producer() may not be None when the value is an input of a GraphView
+
+        # If the value is not in the value map, it may be a graph input or an outer-scope value.
```
titaiwangms: Sorry, I didn't follow Cloner. Can you point out why this statement is valid?

titaiwangms (approved):

src/onnx_ir/_cloner.py
justinchuby: > why this statement is valid?

The value will not be found in two cases: one is when we start the setup and call clone on all the graph inputs, and the second is when we are cloning nodes and see inputs that we have not seen. This is when we know the node input value is not coming from the graph.

src/onnx_ir/_cloner.py
```diff
@@ -170,11 +200,17 @@ def clone_node(self, node: _core.Node) -> _core.Node:
...
     def clone_graph(self, graph: _core.Graph | _core.GraphView) -> _core.Graph:
         """Clones a graph with shared TensorProtocols."""
-        input_values = [self.clone_value(v) for v in graph.inputs]
+        if self._preserve_outer_scope_values and not self._top_scope_local_values:
+            # Collect local values only when we are preserving outer-scope values.
+            # Only set this once for the top-level graph so nested subgraphs can
+            # correctly identify outer-scope values.
+            self._top_scope_local_values = _collect_local_values(graph)
+
+        input_values = [self._clone_value(v) for v in graph.inputs]
```
gramalingam: I think the issue may stem from overloading _clone_value to both clone a value as well as to lookup a clone of a value. But we should know what we want to do at the call-site ... and can potentially call different methods. Here, for example, we want to create a clone (and it should not be in the map already). If we do this, perhaps there is no need to scan the graph and identify local values

justinchuby: Thanks - updated to remove the local value set and changed logic in clone nodes to detect out of scope values. It looks much cleaner now.

src/onnx_ir/_cloner.py
```diff
@@ -58,15 +58,19 @@ def __init__(
                 processing on the cloned node.
             resolve_ref_attrs: Whether to resolve reference attributes using the attr_map.
                 Set to True when inlining functions.
+            preserve_outer_scope_values: Whether to preserve outer-scope values (not defined
```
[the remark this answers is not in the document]
justinchuby: An error works for me - I can create this option instead and default it to true

gramalingam (approved):

PR #314, opened 2026-01-17
Copy node output shape/dtype information in clone()

justinchuby:
- Fix a bug where node outputs do not have shape/dtype information populated
- Improve typing
- Avoid initializing the metadata objects if they are empty in the original value/node/graph

Follow up of https://github.com/onnx/ir-py/pull/313.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,7 +77,7 @@ jobs:
         # To toggle linter comments in the files page, press `i` on the keyboard
         if: always()
         continue-on-error: true
-        uses: github/codeql-action/upload-sarif@v4.31.9
+        uses: github/codeql-action/upload-sarif@v4.31.10
         with:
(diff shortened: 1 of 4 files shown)


src/onnx_ir/_core_test.py
```diff
@@ -3966,6 +3966,38 @@ def test_clone_preserves_value_types_and_shapes(self):
         self.assertEqual(len(cloned_graph.inputs), len(graph.inputs))
         self.assertEqual(len(cloned_graph.outputs), len(graph.outputs))
 
+    def test_clone_preserves_node_output_types_and_shapes(self):
```
titaiwangms: Tests are piling up real quick. We need to keep an eye on the test quality that copilot generated.

enpasos:
[ir-py_clone2](https://github.com/enpasos/jax2onnx/tree/ir-py_clone2) passes tests based on this PR

PR #318, opened 2026-01-20
Fix metadata loss in Graph.clone

enpasos:
The current Cloner implementation creates fresh Value objects for node outputs during cloning, which initializes them with empty attributes. This results in the loss of type, shape, and custom metadata_props from the original graph.

This PR updates _clone_or_get_value and clone_node to explicitly copy these properties to the new values. This ensures that Graph.clone() produces a faithful copy of the input graph, preserving all shape information and custom metadata.

With this PR applied, jax2onnx is able to use the native clone method and pass all its tests (see https://github.com/enpasos/jax2onnx/tree/ir-py_clone).


enpasos:
Sure ... I will test #314

PR #327, opened 2026-01-28
Create the infrastructure for symbolic shape inferencing

justinchuby:
I guided copilot and vibe coded this thing but it looks good.

# Add Symbolic Shape Inference Capability

## Summary

This PR introduces symbolic shape inference capability for the ONNX IR, enabling shape propagation through graphs with support for symbolic dimensions and arithmetic expressions via SymPy.

### Core Enhancements

**`SymbolicDim` class (`src/onnx_ir/_core.py`)**
- Enhanced to support SymPy expressions in addition to string names and `None`
- Added arithmetic operations (`+`, `-`, `*`, `//`, `%`) that produce new symbolic expressions
- Lazy SymPy `Symbol` creation for performance (only created when `.expr` is accessed)
- Pre-computed `value` and `hash` at initialization since `SymbolicDim` is immutable

**`Shape` class (`src/onnx_ir/_core.py`)**
- Added `evaluate(bindings)` method to substitute symbolic dims with concrete values
- Added `simplify()` method to simplify symbolic expressions in dimensions
- Added `free_symbols()` method to get all symbolic variable names in the shape

### New Shape Inference Module (`src/onnx_ir/shape_inference/`)

- **`_registry.py`**: Opset-aware registry for shape inference functions

- **`_context.py`**: `ShapeInferenceContext` for tracking inferred shapes/dtypes
  - Four merge policies: `SKIP`, `OVERRIDE`, `REFINE`, `STRICT`

- **`_broadcast.py`**: NumPy-style broadcasting utilities for shape inference

- **`ops/_add.py`**: Shape and dtype inference for Add operator
- **`ops/_transpose.py`**: Shape and dtype inference for Transpose operator

### New Pass (`src/onnx_ir/passes/common/symbolic_shape_inference.py`)

- `SymbolicShapeInferencePass`: Traverses graph in topological order and infers shapes/dtypes
- `infer_symbolic_shapes()`: Convenience function for one-shot inference

### Dependencies

- Added `sympy>=1.13` to `pyproject.toml`

## Design Decisions

1. **Backward Compatible**: `SymbolicDim` accepts `str | None | sympy.Expr`, maintaining full compatibility with existing code

2. **Lazy Evaluation**: SymPy symbols are only created when needed (accessing `._expr`), avoiding performance overhead for simple cases

3. **Merge Policies**: Different strategies for combining inferred shapes with existing shapes:
   - `SKIP`: Keep existing if present
   - `OVERRIDE`: Always replace
   - `REFINE`: Update only if more specific
   - `STRICT`: Raise on conflicts

4. **Modular Ops**: Each operator's inference logic is in its own file, making it easy to add new operators

5. **Opset Versioning**: Registry supports multiple version specifications for handling operator changes across ONNX opsets

## Testing

Added 50 unit tests covering:
- `SymbolicDim` arithmetic and edge cases
- `Shape.evaluate()`, `simplify()`, `free_symbols()`
- Broadcasting logic
- Registry version handling
- Merge policies
- End-to-end pass execution

## Future Work

- Add more operator implementations
- Constraint system for dimension equality propagation
- Bidirectional inference

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -8,6 +8,7 @@ ir_convenience
 ir_external_data
 ir_passes
 ir_passes_common
+ir_analysis
 ir_traversal
 ir_tape
 ir_tensor_adapters
--- a/docs/api/ir_analysis.md
+++ b/docs/api/ir_analysis.md
@@ -1,0 +1,10 @@
+# onnx_ir.analysis
+
+```{eval-rst}
+.. automodule:: onnx_ir.analysis
+.. currentmodule:: onnx_ir.analysis
+```
+
+```{eval-rst}
+.. autofunction:: analyze_implicit_usage
+```
--- a/src/onnx_ir/__init__.py
+++ b/src/onnx_ir/__init__.py
@@ -174,4 +174,4 @@ def __set_module() -> None:
 
 
 __set_module()
-__version__ = "0.1.16"
+__version__ = "0.2.0"


docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,325 @@
...
+    def __init__(
+        self,
+        model: ir.Model,
+        opset: int | None = None,
+        policy: ShapeMergePolicy = "refine",
+    ) -> None:
+        ...
+
+    def set_shape(self, value: ir.Value, shape: ir.Shape) -> bool: ...
+    def set_dtype(self, value: ir.Value, dtype: ir.DataType) -> bool: ...
```
gramalingam: I am wondering why we have the above two methods? In ONNX itself, an InferenceContext is used because it is intended to work with different IR representations for an onnx model ... and it is used with, eg., both onnxruntime's Graph representation as well as the one inside onnx repo. Here, a value is ir.Value ... so we are working with only this one IR representation. So, do we really need set_shape or set_dtype?

docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,325 @@
...
+    def bind(self, symbol: str, value: int | sympy.Expr) -> None: ...
```
gramalingam: On the other hand, bind makes sense ... because if we bind a symbolic dim "N" to a constant 50 in one place, we can, in principle, update N to 50 everywhere it occurs ... this is something that does not happen with the existing shape-inference. We need some global context to enable something like that.

src/onnx_ir/_core.py
```diff
@@ -1242,47 +1243,229 @@ class SymbolicDim(_protocols.SymbolicDimProtocol, _display.PrettyPrintable):
...
+    @property
+    def expr(self) -> sympy.Expr | None:
+        """The underlying SymPy expression (lazily created).
+
+        Returns the SymPy expression for this dimension, or None for unknown dimensions.
+        """
+        if self._value is None:
+            return None
+        if self._expr is None:
+            self._expr = sympy.Symbol(self._value, integer=True, positive=True)
```
justinchuby: Hide this property. Also how is value parsed? We should support sympy expressions from pytorch, while ensuring security

docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+### 3. Shape Inference Registry
+
+An opset-aware registry maps `(domain, op_type, version)` to inference functions:
+
+```python
+registry = OpShapeInferenceRegistry()
+
+@registry.register("", "Add", versions=1)  # Version 1 and above
```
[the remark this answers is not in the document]
justinchuby: Updated to use since_version

docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,324 @@
...
+
+```python
+batch = ir.SymbolicDim("batch")
+seq = ir.SymbolicDim("seq")
+
+# Create shape with arithmetic
+shape = ir.Shape([batch, seq // 2, 256])
+
+# Evaluate with concrete values
+concrete = shape.evaluate({"batch": 4, "seq": 128})  # [4, 64, 256]
```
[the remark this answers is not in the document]
justinchuby: Should we actually return another ir.Shape?

inisis:
> @inisis could you take a look if this looks good to you or if you can spot any issues? Thanks

@justinchuby I'd love to help, but I think with your domain knowledge and vibe coding, it can not be even better. What I see here is more operator implementations and tests for it.

src/onnx_ir/_core.py
```diff
@@ -1242,47 +1244,237 @@ class SymbolicDim(_protocols.SymbolicDimProtocol, _display.PrettyPrintable):
...
+    def __add__(self, other: int | SymbolicDim) -> SymbolicDim:
+        """Add an integer or another SymbolicDim to this dimension."""
+        if self._expr is None:
+            return SymbolicDim(None)
+        if isinstance(other, int):
+            return SymbolicDim(self._expr + other)
+        if isinstance(other, SymbolicDim):
+            if other._value is None:
+                return SymbolicDim(None)
+            return SymbolicDim(self._expr + other._expr)
```
titaiwangms: Should we embed symplify automatically to save some efforts for the users?

src/onnx_ir/_core.py
```diff
@@ -1242,47 +1244,237 @@ class SymbolicDim(_protocols.SymbolicDimProtocol, _display.PrettyPrintable):
...
+            11
+        """
+        if self._expr is None:
+            return None
+        # Build substitution map using the actual symbols present in the expression
+        subs = {
+            symbol: bindings[str(symbol)]
+            for symbol in self._expr.free_symbols
+            if str(symbol) in bindings
+        }
```
titaiwangms: Do we warn for those not mapped to any symbolic dim?

src/onnx_ir/_core.py
justinchuby: Added calls to sympy.sympify in the magic methods.

src/onnx_ir/_core.py
justinchuby: I updated evaluate to return an IR object. PTAL

src/onnx_ir/passes/common/symbolic_shape_inference_test.py
```diff
@@ -0,0 +1,504 @@
...
+
+        with self.assertRaises(ValueError) as cm:
+            ctx.set_shape(value, ir.Shape([4, 2, 3]))
+        self.assertIn("conflict", str(cm.exception).lower())
+
+
+class SymbolicShapeInferencePassTest(unittest.TestCase):
+    """Tests for SymbolicShapeInferencePass."""
+
+    def test_add_shape_inference(self):
```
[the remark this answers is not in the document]
justinchuby: PTAL

src/onnx_ir/_core.py
```diff
@@ -48,6 +48,7 @@
 
 import ml_dtypes
 import numpy as np
+import sympy
 from typing_extensions import TypeIs, deprecated
```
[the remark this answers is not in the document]
justinchuby: 469609 us of import time. It is ok.

docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,324 @@
...
+An opset-aware registry maps `(domain, op_type, version)` to inference functions with O(1) lookup:
+
+```python
+registry = OpShapeInferenceRegistry()
+
+@registry.register("", "Add", since_version=7)  # Version 7 and above
+def infer_add_v7(ctx, node):
+    ...
+
+@registry.register("", "Reshape", since_version=14)  # Version 14 and above
```
xadupre: Do we need to update the registration everytime we add a new type in onnx specs? (but the shape logic does not change).

docs/design/01_symbolic_shape_inference.md
```diff
@@ -0,0 +1,324 @@
...
+
+### 3. Shape Inference Registry
+
+An opset-aware registry maps `(domain, op_type, version)` to inference functions with O(1) lookup:
+
+```python
+registry = OpShapeInferenceRegistry()
+
+@registry.register("", "Add", since_version=7)  # Version 7 and above
+def infer_add_v7(ctx, node):
```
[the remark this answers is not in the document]
justinchuby: Yes we can register multiple ops. I will do that in a separate PR so that this one focuses on the inference mechanism.

docs/design/01_symbolic_shape_inference.md
justinchuby: I don't think so. It will work for this version and up

src/onnx_ir/shape_inference/_ops/_add.py
```diff
@@ -0,0 +1,52 @@
...
+
+from onnx_ir.shape_inference._broadcast import broadcast_shapes
+from onnx_ir.shape_inference._registry import registry
+
+if TYPE_CHECKING:
+    import onnx_ir as ir
+    from onnx_ir.shape_inference._context import ShapeInferenceContext
+
+
+@registry.register("", "Add", since_version=7)
```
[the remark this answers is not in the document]
justinchuby: Will do it as a follow up.

src/onnx_ir/_symbolic_shapes.py
```diff
@@ -0,0 +1,314 @@
...
+        if char == ",":
+            self.pos += 1
+            return ("COMMA", ",")
+
+        raise ValueError(
+            f"Unexpected character '{char}' at position {self.pos} in expression '{self.text}'"
+        )
+
+
+class _ExpressionParser:
```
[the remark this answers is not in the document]
justinchuby: Let me ask copilot do run a comparison.

PR #343, opened 2026-02-12
Improve magic methods in SymbolicDim

justinchuby:
Implement the missing magic functions, and allow `.` in dim names (seen in some models)


src/onnx_ir/_symbolic_shapes_test.py
```diff
@@ -109,6 +109,42 @@ def test_rmul(self):
...
+
+    def test_floor(self):
+        import math
+
+        dim = ir.SymbolicDim("N")
+        result = math.floor(dim / 3)
+        self.assertEqual(result.value, "floor(N/3)")
+
+    def test_trunc(self):
+        import math
```
justinchuby: Move to top

Issue #333, opened 2026-02-09
Issue when cloning MetadataStore


AyoubMDL:
Hello,
 
When cloning a model whose nodes contain a `MetadataStore`, the underlying metadata objects are shallow copied. Therefore, the original and cloned models share references to the same metadata entries (see code below).
```python
import onnx_ir as ir

def check_meta():
    ir_model = ir.from_onnx_text("""
                < ir_version: 10, opset_import: ["" : 21] >
                test_model (float[N, 32] X) => (float [N, ?] Y)
                {
                    Y = Relu(X)
                }
            """)

    # Add metadata to the node
    ir_model.graph[0].meta["test_meta"] = "test_value"

    # Clone the model
    cloned_ir_model = ir_model.clone()

    # Check that the metadata is cloned
    assert cloned_ir_model.graph[0].meta is not ir_model.graph[0].meta

    # Check that the metadata value is the same
    print(
        cloned_ir_model.graph[0].meta["test_meta"] is ir_model.graph[0].meta["test_meta"]
    )  # -> True

check_meta()
```
The fix is that `MetadataStore` content should be deep copied. I can open a PR if necessary.

AyoubMDL:
> Also in your use case, what is meta["input"]? Is it an ir.Value?

It's a numpy array. Also, I use meta to store other things during the process but with no cloning (expect in this unit test)

PR #354, opened 2026-02-19
Optional MetadataStore deep copy when cloning

AyoubMDL:
This PR introduces an optional deep copy when cloning for objects stored in the metadata store (addresses https://github.com/onnx/ir-py/issues/333).

### DIFF
--- a/src/onnx_ir/_metadata.py
+++ b/src/onnx_ir/_metadata.py
@@ -43,3 +43,6 @@ class MetadataStore(collections.UserDict):
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self.data!r}, invalid_keys={self._invalid_keys!r})"
+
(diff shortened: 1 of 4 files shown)


src/onnx_ir/_metadata.py
```diff
@@ -43,3 +43,6 @@ def is_valid(self, key: str) -> bool:
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self.data!r}, invalid_keys={self._invalid_keys!r})"
+
+    def __bool__(self) -> bool:
+        return bool(self.data) or bool(self._invalid_keys)
```
justinchuby: If self.data is empty, invalid keys shouldn't matter?

justinchuby (approved):

justinchuby (approved):

PR #369, opened 2026-03-13
Fix initializer-safe bulk renaming and NameFixPass initializer iteration

### LINKED ISSUE #353: Issue with renaming
1) Direct rename loops fail for initializer-backed outputs (swap/permutation case).
   - Minimal repro: export a model whose two outputs are constants, then try to swap their names in-place.
   - Error:
     ValueError: Cannot rename initializer ... to 'const_1': an initializer with that name already exists.
   - This behavior comes from the initializer guard in Value.name setter:
     onnx_ir/_core.py:2514
     onnx_ir/_core.py:2522

2) “rename first + NameFixPass” can crash on initializer collisions.
   - Minimal repro: rename an input to an existing initializer name, then run NameFixPass.
   - Error:
     RuntimeError: dictionary keys changed during iteration
   - NameFixPass iterates `graph_like.initializers.values()`:
    onnx_ir/passes/common/naming.py:137
   - Renaming an initializer mutates the initializer dict keys:
     onnx_ir/_core.py:2541

Reported in https://github.com/enpasos/jax2onnx/pull/186#issuecomment-3909316043

enpasos:
## Summary

Fixes onnx/ir-py#353.

This PR addresses two related rename failures:

1. Direct rename loops fail for initializer-backed values in swap/permutation cases.
2. `NameFixPass` can raise `RuntimeError: dictionary keys changed during iteration` when a rename changes initializer dict keys while the pass is iterating them.

## Changes

- add `onnx_ir.convenience.rename_values()` for bulk value renaming
- make initializer-backed multi-value renames safe by using temporary names during swaps/permutations
- keep the existing initializer collision checks for names outside the rename set
- make `NameFixPass` iterate over a snapshot of initializer values instead of the live dict view
- add regression tests for:
  - swapping two initializer-backed outputs
  - renaming to an existing initializer name before running `NameFixPass`

## Why

A single `value.name = ...` rename is not enough for safe initializer swaps, because the initializer name guard correctly rejects taking another live initializer name. The new bulk helper provides the missing multi-rename operation, while the `NameFixPass` change makes renaming safe when initializer dict keys are updated during the pass.

## Validation

- `pytest -q src/onnx_ir/_convenience/_convenience_test.py -k rename_values`
- `pytest -q src/onnx_ir/passes/common/naming_test.py -k initializer_collision`

Reported from downstream integration work in jax2onnx.

### DIFF
--- a/src/onnx_ir/_convenience/__init__.py
+++ b/src/onnx_ir/_convenience/__init__.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 __all__ = [
     "convert_attribute",
     "convert_attributes",
+    "rename_values",
     "replace_all_uses_with",
     "create_value_mapping",
     "replace_nodes_and_values",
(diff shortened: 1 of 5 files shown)


src/onnx_ir/_convenience/__init__.py
```diff
@@ -360,6 +361,102 @@ def replace_all_uses_with(
...
+        names = (names,)
+    if len(values) != len(names):
+        raise ValueError("The number of values and names must match.")
+
+    ordered_pairs: list[tuple[_core.Value, str | None]] = []
+    target_by_value_id: dict[int, str | None] = {}
+    for value, name in zip(values, names):
+        if not isinstance(value, _core.Value):
+            raise TypeError(f"value must be a Value object, not {type(value)}")
+        value_id = id(value)
```
justinchuby: no need to use id. values are hashable

justinchuby (commented):
To rename initializers, it is possible to simply pop the initializers, rename, and put them back. Renaming other values is quite straightforward (just set the name field). I think the logic can be simplified.

justinchuby (approved):

justinchuby:
@enpasos could you fix lint?

PR #407, opened 2026-05-08
Improve test coverage

justinchuby:
This pull request adds comprehensive new test coverage for several internal modules in the `onnx_ir` package, focusing on containers, metadata, version utilities, tape/builder logic, and linked list/graph access. The new tests improve reliability by verifying edge cases, error handling, and expected behaviors.

### DIFF
--- a/src/onnx_ir/_display_test.py
+++ b/src/onnx_ir/_display_test.py
@@ -4,6 +4,7 @@
 
 import contextlib
 import io
+import types
 import unittest
 from unittest import mock
 
@@ -49,7 +50,21 @@ class DisplayRichFallbackTest(unittest.TestCase):
         """Test display with page=True uses rich pager."""
         graph = ir.Graph([], [], nodes=[], name="test_graph")
         mock_console = mock.MagicMock()
-        with mock.patch("rich.console.Console", return_value=mock_console):
+        mock_console.pager.return_value = contextlib.nullcontext()
+        fake_rich = types.ModuleType("rich")
+        fake_rich.print = mock.MagicMock()
+        fake_rich.console = types.SimpleNamespace(
+            Console=mock.MagicMock(return_value=mock_console)
+        )
+        fake_rich.markup = types.SimpleNamespace(escape=lambda text: text)
+        with mock.patch.dict(
+            "sys.modules",
+            {
+                "rich": fake_rich,
+                "rich.markup": fake_rich.markup,
+                "rich.console": fake_rich.console,
+            },
+        ):
             graph.display(page=True)
         mock_console.pager.assert_called_once()
         mock_console.print.assert_called_once()
--- a/src/onnx_ir/_display_test.py
+++ b/src/onnx_ir/_display_test.py
@@ -24,5 +24,42 @@ class DisplayTest(unittest.TestCase):
             graph.display()
 
 
+class DisplayRichFallbackTest(unittest.TestCase):
+    def test_require_rich_returns_none_when_not_installed(self):
+        from unittest import mock
+
+        from onnx_ir import _display
+
+        with mock.patch.dict("sys.modules", {"rich": None}):
+            result = _display.require_rich()
+            self.assertIsNone(result)
+
+    def test_display_without_rich_prints_plaintext(self):
+        """Test the fallback when rich is not available."""
+        import io
+        from unittest import mock
+
+        from onnx_ir import _display
+
+        graph = ir.Graph([], [], nodes=[], name="test_graph")
+
+        # Mock require_rich to return None (simulating rich not installed)
+        with mock.patch.object(_display, "require_rich", return_value=None):
+            output = io.StringIO()
+            with contextlib.redirect_stdout(output):
+                graph.display()
+            printed = output.getvalue()
+            self.assertIn("Tip:", printed)
+
+    def test_display_with_page(self):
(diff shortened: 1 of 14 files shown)


titaiwangms (commented):
## Review summary

**QA:** ✅ all 591 new tests pass in 0.89s, zero failures, zero flakiness.
**Verdict:** request changes — mechanically clean, but several tests don't actually verify behavior, and a few look like coverage theater for real bugs in the code under test.

### In-scope issues (tests in this PR)

1. **`_display_test.py:60`** — `test_display_with_page` wraps `graph.display(page=True)` in `contextlib.suppress(Exception)`. The test passes even if `display(page=True)` crashes. Mock the rich pager path and assert it gets invoked.
2. **`_graph_containers_test.py:33`** — `test_graph_initializers_set_and_delete` deletes the initializer and then asserts nothing (the comment says "graph should be unset" but there's no follow-up assert). Add `assertNotIn("w", graph.initializers)` and an ownership-flag check after the delete.
3. **`serde_test.py:975, 1032`** — Tautological for the feature being tested:
   - `assertIsNotNone(result_proto)` after `serialize_model(...)`
   - re-checking `result_proto.ir_version == 9` rather than the function value_info behavior the test name advertises
   Assert concrete `functions[0].value_info` entries (names / types / shapes) for the IR10 case, and the `pkg::my_func/...` value_info entries in the main graph for the IR9 experimental case.
4. **`serde_test.py:32-57`** — `test_from_proto` / `test_to_proto` call dispatchers with no assertions. Wrong return types or lossy serialization would still pass. Either assert concrete return type + at least one meaningful field per case, or drop these in favor of the stronger dispatch tests later in the file.
5. **`_pass_infra_test.py:428-440`** — The test accepts *either* `PostconditionError` *or* `TypeError` when `call()` returns a non-`PassResult`. That pins broken control flow in `_pass_infra.py:129-140`: `ensures(result.model)` runs before validating that `result` is a `PassResult`. Postcondition errors should not be the surface for "your pass returned the wrong type". Tighten the test to require `TypeError`, and move the `isinstance(result, PassResult)` check before `self.ensures(result.model)`.
6. **`_version_utils_test.py:11-22`** — Compares against the installed live `onnx`/`onnxruntime` versions with absurd bounds (`0.0.1`, `99.99.99`). It won't catch the prerelease-ignoring behavior in `_version_utils.py` (which compares only `.release`, so `2.0.0rc1 >= 2.0.0`), and the bounds will eventually become wrong. Monkeypatch `__version__` and cover exact, prerelease, dev, and local-version cases.
7. **`_linked_list_test.py:404`** — `test_getitem_slice` only asserts `len(sliced) == 2`; wrong contents or order would still pass. Use `assertEqual(list(sliced), [n1, n2])`.
8. **`_enums_test.py:157-168`** — `_FLOAT_TYPES` omits `FLOAT8E8M0`, even though `DataType.is_floating_point()` includes it and `ml_dtypes.finfo()` supports it. The finfo-backed properties aren't tested for one of the supported floating types. Add it and make classification exhaustive against the enum.
9. **`serde_test.py:443`** — `test_round_trip_numpy_conversion_from_raw_data` casts NaN/inf into integer dtypes and emits 8 `RuntimeWarning`s. Benign today, but CI-hostile under `-W error`. Split float-special cases from integer cases, or `filterwarnings` the expected warning.

### Production bugs surfaced (out of scope for this PR — worth a follow-up)

While reading the modules under test, the adversarial pass turned up what look like real bugs that the new tests should arguably have caught:

- **`_graph_containers.py:104-110`** — Slice assignment with a generator double-consumes it: `_set_graph()` marks values graph-owned, then `super().__setitem__` receives an exhausted iterator and inserts nothing. Repro:
  ```python
  v1 = ir.Value(name="x"); v2 = ir.Value(name="y")
  g = ir.Graph([v1], [], nodes=[])
  g.inputs[0:1] = (v for v in [v2])
  assert len(g.inputs) == 0          # actually 0
  assert v2._graph is g              # but ownership got set
  ```
  Fix: materialize `item = tuple(item)` before iterating in `__setitem__`.
- **`traversal.py:85-109`** — `RecursiveGraphIterator` fires `enter_graph` / `exit_graph` twice per subgraph: once manually in `_iterate_subgraphs`, then again inside the nested `RecursiveGraphIterator`. Scope-sensitive passes will see duplicate enter/exit events. Either drop the manual callbacks around the nested iterator or disable callbacks in the nested one — and add a callback-order test.
- **`_pass_infra.py:129-140`** — see #5 above; check `isinstance(result, PassResult)` before running postconditions on `result.model`.
- **`_version_utils.py:15-18, 45-48`** — see #6 above; comparison drops prerelease/dev/local segments.

### Praise

- `_linked_list_test.py` has genuinely useful mutation-during-iteration coverage — that's the right structural-testing instinct for that container.
- `unused_removal_test.py` covers nested-graph removal and optional-output trimming with real assertions on high-risk integration paths.
- `serde_test.py` round-trips real protos through several complex features rather than only mocking the dispatch boundary.

Net: nice breadth bump. Tightening the few non-verifying tests above (#1–#5 especially) before merging would mean this batch actually catches the regressions it implies.

titaiwangms (approved):
CI failures

justinchuby:
@copilot please fix CI. You can install `rich` in the test env

PR #417, opened 2026-05-19
Add tensor-centric accessors to GraphInitializers

justinchuby:
## Motivation

`graph.initializers` is a `dict[str, Value]` for backward compatibility, but the value almost every caller actually wants is the backing `Tensor` (`value.const_value`). Reading initializer tensors today requires either

```python
for v in g.initializers.values():
    t = v.const_value     # might also be None
    ...
```

or, per name, `g.initializers[name].const_value` — three layers of indirection for a very common operation. Writing isn't much fun either: you have to construct a `Value(name=..., shape=..., type=..., const_value=tensor)` manually before calling `.add()`.

## What's in this PR (additive only)

Three new methods on `GraphInitializers`:

| method | mirrors | returns |
|---|---|---|
| `tensors()` | `dict.values()` | `Iterator[Tensor]` |
| `tensor_items()` | `dict.items()` | `Iterator[(str, Tensor)]` |
| `get_tensor(name, default=None)` | `dict.get` | `Tensor \| None` |

Typical use:

```python
# Before
for v in g.initializers.values():
    arr = v.const_value.numpy()
    ...

# After
for name, tensor in g.initializers.tensor_items():
    arr = tensor.numpy()
    ...
```

## Design decisions worth confirming

1. **`tensors()` / `tensor_items()` raise on missing `const_value`.** An initializer without a backing tensor is almost always a graph-construction bug — the whole point of an initializer is to ship a tensor. Surfacing that loudly here matches `Mapping[str, Tensor]` semantics. Callers who legitimately need to handle pending values can keep using `.values()` / `.items()` directly.

2. **`get_tensor` is more lenient.** Returns `default` both when the key is missing and when `const_value` is `None`. This matches the spirit of `dict.get` (no exceptions).

3. **No new write API in this PR.** I considered adding `set_tensor(name, tensor)` to remove the verbose `Value(...)` construction for the create-or-update case, and a `as_tensors` read-only `Mapping[str, Tensor]` view that turns the whole `Mapping[str, Tensor]` model into a first-class thing. Both could be follow-ups — I'd like to see if the three read accessors here cover the bulk of the ergonomics first.

## Compatibility

Strictly additive. `graph.initializers[name]` still returns `Value`. `.values()`, `.items()`, `.keys()`, `__iter__`, `__contains__`, `__len__`, `pop`, `update`, `add` — all unchanged. Existing tests (3157 unit tests under `src/` + `tests/`) all pass.

## Tests

`GraphContainersTest` gains six new cases covering happy path, name+tensor pairs, default handling, and the missing-`const_value` raise behaviour.

### DIFF
--- a/src/onnx_ir/_core_test.py
+++ b/src/onnx_ir/_core_test.py
@@ -3288,6 +3288,58 @@ class GraphContainersTest(unittest.TestCase):
         self.assertTrue(initializers[0].is_initializer())
         self.assertEqual(initializers[0].graph, self.graph)
 
+    def test_tensors_yields_const_values(self):
+        self.graph.initializers.add(self.value3)
+        other = _core.Value(name="initializer2", const_value=ir.tensor([4, 5]))
+        self.graph.initializers.add(other)
+
+        tensors = list(self.graph.initializers.tensors())
+        self.assertEqual(len(tensors), 2)
+        self.assertIs(tensors[0], self.value3.const_value)
+        self.assertIs(tensors[1], other.const_value)
+
+    def test_tensor_items_yields_name_tensor_pairs(self):
+        self.graph.initializers.add(self.value3)
+
+        items = list(self.graph.initializers.tensor_items())
+        self.assertEqual(len(items), 1)
+        name, tensor = items[0]
+        self.assertEqual(name, "initializer1")
+        self.assertIs(tensor, self.value3.const_value)
+
+    def test_tensors_raises_when_const_value_missing(self):
+        pending = _core.Value(name="pending")  # no const_value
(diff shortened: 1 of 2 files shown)


justinchuby:
@gramalingam do these APIs look good to you?

cyyever (approved):

justinchuby:
@copilot I think we also need a way to iterate on initializer values. do you think initializers.values() is good enough or are there better ways?

gramalingam (approved):

justinchuby:
@copilot investigate how many call sites can be simplified with the new APIs. Just give me an idea without any code changes.

PR #423, opened 2026-05-22
Use new GraphInitializers tensor APIs to simplify codebase

justinchuby:
Accesses to initializer tensors across the test suite still used the old three-layer pattern (`initializers["name"].const_value`, `for value in .values(): value.const_value`) introduced before the `get_tensor()` / `tensors()` / `tensor_items()` APIs existed.

## Changes

- **`initializers["name"].const_value` → `initializers.get_tensor("name")`** — applied across `external_data_test.py`, `serde_test.py`, `_safetensors_test.py`, `_io_test.py`, `shape_inference_test.py`, `initializer_deduplication_test.py`, `constant_manipulation_test.py`
- **`for value in .values(): value.const_value` → `for tensor in .tensors()`** — where only the tensor is needed
- **`for name, value in .items(): value.const_value` → `for name, tensor in .tensor_items()`** — where name + tensor are both needed

```python
# Before
external_tensor = model.graph.initializers["tensor1"].const_value
for value in model.graph.initializers.values():
    self.assertIsInstance(value.const_value, ir.Tensor)

# After
external_tensor = model.graph.initializers.get_tensor("tensor1")
for tensor in model.graph.initializers.tensors():
    self.assertIsInstance(tensor, ir.Tensor)
```

Production code is left unchanged: those loops either require the `Value` object for operations unrelated to the tensor (naming, `replace_all_uses_with`, etc.) or deliberately skip `None` const_values — behaviour that differs from `tensors()`'s strict-raise semantics.

### DIFF
--- a/src/onnx_ir/_io_test.py
+++ b/src/onnx_ir/_io_test.py
@@ -60,7 +60,7 @@ class IOFunctionsTest(unittest.TestCase):
         self.assertEqual(len(loaded_model.graph.initializers), 1)
         self.assertEqual(len(loaded_model.graph), 2)
         np.testing.assert_array_equal(
-            loaded_model.graph.initializers["initializer_0"].const_value.numpy(),
+            loaded_model.graph.initializers.get_tensor("initializer_0").numpy(),
             np.array([0.0]),
         )
         np.testing.assert_array_equal(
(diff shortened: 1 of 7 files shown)


justinchuby:
These are all tests. So I don't know how useful the API is, especially when the setter is not mirroring it. @gramalingam do you have thoughts?

PR #427, opened 2026-05-22
Support sharding plain ONNX external data

### LINKED ISSUE #421: Support sharding onnx_data
Right now, we support sharding when safetensors is selected as the external data format. Safetensors has limitations when used with ORT cuda because we do not control the alignment of the tensors, which makes them incompatible with the cuda kernels sometimes.

We want to be able to shard plain onnx data such that sharding can be used without safetensors.

justinchuby:
## Summary

Adds support for **sharding plain ONNX external data** across multiple `.data` files. A new `max_shard_size_bytes` parameter on `ir.save()` / `external_data.unload_from_model()` distributes a model's tensors across numbered shard files (e.g. `model-00001-of-00003.data`) so that no single file exceeds the configured budget.

Unlike safetensors, **no separate index file is needed**: the ONNX proto already stores `location`, `offset`, and `length` per tensor, so the `.onnx` itself encodes which shard each tensor lives in.

## Motivation

Safetensors sharding is incompatible with ORT CUDA kernels due to uncontrolled tensor alignment. Plain ONNX external data preserves the alignment guarantees ORT needs, but until now was capped at a single data file — which becomes unwieldy for larger models. Sharding lets large models stay on the plain-ONNX external-data format while still being splittable.

## Changes

- **`external_data.py`**
  - `_get_shard_filename()` — generates numbered shard filenames (e.g. `model-00001-of-00003.data`); returns the original name when `total_shards == 1`. Uses `os.path.splitext` so paths with a dotted directory (e.g. `my.dir/model`) are handled correctly.
  - `_shard_tensors()` — greedily partitions tensors into shard groups respecting `max_shard_size_bytes`. Uses an O(1)-per-tensor `_ShardSizeAccumulator` that tracks the on-disk size (alignment-aware) without re-sorting.
  - `unload_from_model()` — new `max_shard_size_bytes: int | None = None` kwarg; when set, distributes tensors across multiple shard files. Callback `total`/`index` reflect global position across shards. Raises `FileExistsError` if the sharded path would clobber a pre-existing destination file that the in-memory model doesn't own (avoids silently corrupting another model's data sitting in the same directory). Tensors larger than `max_shard_size_bytes` are written in their own oversized shard with a warning.
  - `_check_no_foreign_shard_collisions()` / `_collect_owned_external_files()` — the simple collision check that backs the new safety contract.
- **`_io.py`**
  - `save()` — new `max_shard_size_bytes` kwarg threaded through to `unload_from_model()`. Validates that `max_shard_size_bytes > 0` and that it isn't set without `external_data`.

## Usage

```python
import onnx_ir as ir

model = ir.load("model.onnx")

# Write tensors ≥ 256 bytes into ≤ 5 GB shard files
ir.save(
    model,
    "model.onnx",
    external_data="model.data",        # base name; shards become model-00001-of-00003.data etc.
    size_threshold_bytes=256,
    max_shard_size_bytes=5 * 1000**3,  # 5 GB per shard
)
```

## Safety contract for sharded saves

- **Self-overwrite of a loaded model** (`ir.load(path)` → re-save to the same `path`) just works: the loaded `ExternalTensor`s are recognized as owned and overwriting them is allowed.
- **Foreign collision** (a file at a target shard path that the model doesn't reference) → `FileExistsError`. Caller must delete the conflicting files or save into a different directory.
- **Unrelated shard-shaped files** in the same directory that *don't* collide with the new shard filenames are left alone.
- **No automatic cleanup** of stale shards from a prior layout (deliberately — that path led to a cross-model deletion bug). If the layout changes (e.g. 4 shards → 2), old shards from the previous save remain on disk for the caller to clean up.

## Tests

- Unit tests for `_get_shard_filename`, `_shard_tensors`, and `_ShardSizeAccumulator` (200k-trial accumulator fuzz, uint8 boundary tests at `_ALIGN_THRESHOLD ± 1`, shuffled-add-order invariance).
- Integration tests for `_io.save` covering single-shard, multi-shard, oversized-tensor warning, load → re-save with the same and different shard counts, sharded → non-sharded transition, foreign-collision raise, and unrelated-file isolation.

Full suite passes (`1628 tests, 8 subtests`).

### DIFF
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -99,17 +99,24 @@ def save(
             files (e.g. ``model-00001-of-00003.data``). Because the ONNX format stores
             ``location``, ``offset``, and ``length`` per tensor, no separate index file is
             created — the saved ONNX proto itself encodes which shard each tensor lives in.
+            If a single tensor is larger than this value, it is written in its own oversized
+            shard file.
             Effective only when ``external_data`` is set.
         callback: A callback function that is called for each tensor that is saved to external data
             for debugging or logging purposes.
 
     Raises:
         ValueError: If the external data path is an absolute path.
+        ValueError: If ``max_shard_size_bytes`` is not greater than 0.
     """
     if external_data is not None:
         if os.path.isabs(external_data):
             raise ValueError(
                 f"The external data path must be relative to the ONNX file path, not '{external_data}'."
+            )
+        if max_shard_size_bytes is not None and max_shard_size_bytes <= 0:
+            raise ValueError(
+                f"max_shard_size_bytes must be greater than 0, got {max_shard_size_bytes}."
             )
         base_dir = os.path.dirname(path)
 
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -110,14 +110,11 @@ def save(
         ValueError: If ``max_shard_size_bytes`` is not greater than 0.
         ValueError: If ``max_shard_size_bytes`` is set without ``external_data``.
         FileExistsError: When ``max_shard_size_bytes`` is set and any
-            destination shard file already exists on disk but is not
-            referenced by an :class:`ExternalTensor` in the model being
-            saved. TODO(justinchuby): the single-file (``max_shard_size_bytes
-            is None``) write path is currently permissive and silently
-            overwrites whatever is at ``external_data``; the collision check
-            is sharded-only because shard layouts are ambiguous (different
-            shard counts produce different filenames). The two paths should
-            eventually share the same overwrite policy.
+            destination shard file already exists on disk. The sharded write
+            path never overwrites existing files; delete the conflicting
+            files or choose a different external data path to re-save. The
+            single-file path (``max_shard_size_bytes is None``) instead
+            overwrites ``external_data`` unconditionally and never raises here.
     """
     if max_shard_size_bytes is not None and external_data is None:
         raise ValueError(
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -16,6 +16,49 @@ from onnx_ir import external_data as _external_data
 from onnx_ir._polyfill import zip
 
 
+def _read_existing_external_data_locations(
+    model_path: str | os.PathLike, base_dir: str | os.PathLike
+) -> list[str]:
+    """Return the absolute external_data file paths referenced by an existing ``.onnx``.
+
+    Used by :func:`save` to recover the set of files written by the previous
+    save of *this* model. The returned paths feed into
+    :func:`external_data.unload_from_model`'s ownership-scoped cleanup so that
+    stale shard files from a previous layout are removed without ever touching
+    shard-shaped files that happen to belong to a different model in the same
+    directory.
+
+    Returns an empty list if ``model_path`` does not exist, is not readable as
+    an ONNX proto, or contains no external initializers.
+    """
+    if not os.path.exists(model_path):
+        return []
+    try:
+        proto = onnx.load(model_path, load_external_data=False)
+    except Exception:  # pragma: no cover - best-effort recovery
+        return []
+    base_dir_str = os.fspath(base_dir) if base_dir else "."
+    locations: set[str] = set()
+
+    def _harvest(graph_proto) -> None:
+        for init in graph_proto.initializer:
+            if init.data_location != onnx.TensorProto.EXTERNAL:
+                continue
+            for entry in init.external_data:
+                if entry.key == "location":
+                    locations.add(os.path.realpath(os.path.join(base_dir_str, entry.value)))
+        for node in graph_proto.node:
+            for attr in node.attribute:
+                if attr.type == onnx.AttributeProto.GRAPH:
+                    _harvest(attr.g)
+                elif attr.type == onnx.AttributeProto.GRAPHS:
+                    for sub in attr.graphs:
+                        _harvest(sub)
+
+    _harvest(proto.graph)
+    return sorted(locations)
+
+
 def load(path: str | os.PathLike, format: str | None = None) -> _core.Model:
     """Load an ONNX model from a file.
 
@@ -126,6 +169,16 @@ def save(
             )
         base_dir = os.path.dirname(path)
 
+        # Look at the existing .onnx file at ``path`` (if any) and collect the
+        # external_data ``location`` entries it references. These count as
+        # files owned by *this* model for cleanup purposes, even when the
+        # in-memory ``model`` was constructed in-memory and has no
+        # ExternalTensor initializers to learn ownership from. This is what
+        # lets re-saving with a different shard layout (e.g. sharded → single
+        # file, or N shards → M shards) safely clean up stale shards without
+        # touching unrelated models' files sitting in the same directory.
+        prior_external_files = _read_existing_external_data_locations(path, base_dir)
+
         # Store the original initializer values so they can be restored if modify_model=False
         initialized_values: list[_core.Value] = []
         for graph in model.graphs():
@@ -141,6 +194,7 @@ def save(
                 size_threshold_bytes=size_threshold_bytes,
                 max_shard_size_bytes=max_shard_size_bytes,
                 callback=callback,
+                prior_external_files=prior_external_files,
             )
             proto = serde.serialize_model(model)
             onnx.save(proto, path, format=format)
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -44,6 +44,7 @@ def save(
     format: str | None = None,
     external_data: str | os.PathLike | None = None,
     size_threshold_bytes: int = 256,
+    max_shard_size_bytes: int | None = None,
     callback: Callable[[_protocols.TensorProtocol, _external_data.CallbackInfo], None]
     | None = None,
 ) -> None:
@@ -92,6 +93,13 @@ def save(
             it will be serialized in the ONNX Proto message.
         size_threshold_bytes: Save to external data if the tensor size in bytes is larger than this threshold.
             Effective only when ``external_data`` is set.
+        max_shard_size_bytes: Maximum cumulative size in bytes for a single external data shard file.
+            When ``None`` (the default) all external tensors are written to the single file
+            given by ``external_data``. When set, tensors are distributed across numbered shard
+            files (e.g. ``model-00001-of-00003.data``). Because the ONNX format stores
+            ``location``, ``offset``, and ``length`` per tensor, no separate index file is
+            created — the saved ONNX proto itself encodes which shard each tensor lives in.
+            Effective only when ``external_data`` is set.
         callback: A callback function that is called for each tensor that is saved to external data
             for debugging or logging purposes.
 
@@ -118,6 +126,7 @@ def save(
                 base_dir,
                 external_data,
                 size_threshold_bytes=size_threshold_bytes,
+                max_shard_size_bytes=max_shard_size_bytes,
                 callback=callback,
             )
             proto = serde.serialize_model(model)
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -108,7 +108,13 @@ def save(
     Raises:
         ValueError: If the external data path is an absolute path.
         ValueError: If ``max_shard_size_bytes`` is not greater than 0.
+        ValueError: If ``max_shard_size_bytes`` is set without ``external_data``.
     """
+    if max_shard_size_bytes is not None and external_data is None:
+        raise ValueError(
+            "max_shard_size_bytes can only be used together with external_data; "
+            "set external_data to the relative path where shards should be written."
+        )
     if external_data is not None:
         if os.path.isabs(external_data):
             raise ValueError(
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -109,6 +109,15 @@ def save(
         ValueError: If the external data path is an absolute path.
         ValueError: If ``max_shard_size_bytes`` is not greater than 0.
         ValueError: If ``max_shard_size_bytes`` is set without ``external_data``.
+        FileExistsError: When ``max_shard_size_bytes`` is set and any
+            destination shard file already exists on disk but is not
+            referenced by an :class:`ExternalTensor` in the model being
+            saved. TODO(justinchuby): the single-file (``max_shard_size_bytes
+            is None``) write path is currently permissive and silently
+            overwrites whatever is at ``external_data``; the collision check
+            is sharded-only because shard layouts are ambiguous (different
+            shard counts produce different filenames). The two paths should
+            eventually share the same overwrite policy.
     """
     if max_shard_size_bytes is not None and external_data is None:
         raise ValueError(
--- a/src/onnx_ir/_io.py
+++ b/src/onnx_ir/_io.py
@@ -16,49 +16,6 @@ from onnx_ir import external_data as _external_data
 from onnx_ir._polyfill import zip
 
 
-def _read_existing_external_data_locations(
-    model_path: str | os.PathLike, base_dir: str | os.PathLike
-) -> list[str]:
-    """Return the absolute external_data file paths referenced by an existing ``.onnx``.
-
-    Used by :func:`save` to recover the set of files written by the previous
-    save of *this* model. The returned paths feed into
-    :func:`external_data.unload_from_model`'s ownership-scoped cleanup so that
-    stale shard files from a previous layout are removed without ever touching
-    shard-shaped files that happen to belong to a different model in the same
-    directory.
-
-    Returns an empty list if ``model_path`` does not exist, is not readable as
-    an ONNX proto, or contains no external initializers.
-    """
-    if not os.path.exists(model_path):
-        return []
-    try:
-        proto = onnx.load(model_path, load_external_data=False)
-    except Exception:  # pragma: no cover - best-effort recovery
-        return []
-    base_dir_str = os.fspath(base_dir) if base_dir else "."
-    locations: set[str] = set()
-
-    def _harvest(graph_proto) -> None:
-        for init in graph_proto.initializer:
-            if init.data_location != onnx.TensorProto.EXTERNAL:
-                continue
-            for entry in init.external_data:
-                if entry.key == "location":
-                    locations.add(os.path.realpath(os.path.join(base_dir_str, entry.value)))
-        for node in graph_proto.node:
-            for attr in node.attribute:
-                if attr.type == onnx.AttributeProto.GRAPH:
-                    _harvest(attr.g)
-                elif attr.type == onnx.AttributeProto.GRAPHS:
-                    for sub in attr.graphs:
-                        _harvest(sub)
-
-    _harvest(proto.graph)
-    return sorted(locations)
-
-
 def load(path: str | os.PathLike, format: str | None = None) -> _core.Model:
     """Load an ONNX model from a file.
 
@@ -169,16 +126,6 @@ def save(
             )
         base_dir = os.path.dirname(path)
 
-        # Look at the existing .onnx file at ``path`` (if any) and collect the
-        # external_data ``location`` entries it references. These count as
-        # files owned by *this* model for cleanup purposes, even when the
-        # in-memory ``model`` was constructed in-memory and has no
-        # ExternalTensor initializers to learn ownership from. This is what
-        # lets re-saving with a different shard layout (e.g. sharded → single
-        # file, or N shards → M shards) safely clean up stale shards without
-        # touching unrelated models' files sitting in the same directory.
-        prior_external_files = _read_existing_external_data_locations(path, base_dir)
-
         # Store the original initializer values so they can be restored if modify_model=False
         initialized_values: list[_core.Value] = []
         for graph in model.graphs():
@@ -194,7 +141,6 @@ def save(
                 size_threshold_bytes=size_threshold_bytes,
                 max_shard_size_bytes=max_shard_size_bytes,
                 callback=callback,
-                prior_external_files=prior_external_files,
             )
             proto = serde.serialize_model(model)
             onnx.save(proto, path, format=format)
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -913,6 +913,12 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
             chunk_size = 1024 * 1024  # 1MB
             while bytes_to_copy > 0:
                 chunk = src.read(min(chunk_size, bytes_to_copy))
+                if not chunk:
+                    raise OSError(
+                        f"External data file {self.path!r} is shorter than expected: "
+                        f"could not read {bytes_to_copy} more byte(s) at offset "
+                        f"{(self._offset or 0) + ((self._length or self.nbytes) - bytes_to_copy)}."
+                    )
                 file.write(chunk)
                 bytes_to_copy -= len(chunk)
 
--- a/src/onnx_ir/_core.py
+++ b/src/onnx_ir/_core.py
@@ -914,10 +914,13 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
             while bytes_to_copy > 0:
                 chunk = src.read(min(chunk_size, bytes_to_copy))
                 if not chunk:
+                    failed_at_offset = (self._offset or 0) + (
+                        (self._length or self.nbytes) - bytes_to_copy
+                    )
                     raise OSError(
                         f"External data file {self.path!r} is shorter than expected: "
                         f"could not read {bytes_to_copy} more byte(s) at offset "
-                        f"{(self._offset or 0) + ((self._length or self.nbytes) - bytes_to_copy)}."
+                        f"{failed_at_offset}."
                     )
                 file.write(chunk)
                 bytes_to_copy -= len(chunk)
--- a/src/onnx_ir/_core_test.py
+++ b/src/onnx_ir/_core_test.py
@@ -510,6 +510,33 @@ class ExternalTensorTest(unittest.TestCase):
         # Ensure repeated reads are consistent
         np.testing.assert_equal(tensor, self.data)
 
+    def test_tofile_raises_oserror_when_backing_file_is_truncated(self):
+        # tofile copies ``length`` bytes from the backing file in chunks. If
+        # the file is shorter than expected (e.g. the data file was truncated
+        # or replaced) the loop must raise an actionable OSError instead of
+        # spinning forever on empty reads.
+        external_data_name = "short.bin"
+        data = np.arange(64, dtype=np.float32)
+        full_bytes = data.tobytes()
+        data_path = pathlib.Path(self.base_path) / external_data_name
+        with open(data_path, "wb") as f:
+            f.write(full_bytes)
+        tensor = _core.ExternalTensor(
+            external_data_name,
+            offset=0,
+            length=len(full_bytes),
+            dtype=ir.DataType.FLOAT,
+            base_dir=self.base_path,
+            name="w",
+            shape=_core.Shape(data.shape),
+        )
+        # Truncate the backing file so it is shorter than ``length``.
+        with open(data_path, "wb") as f:
+            f.write(full_bytes[: len(full_bytes) // 2])
+        buffer = io.BytesIO()
+        with self.assertRaisesRegex(OSError, "shorter than expected"):
+            tensor.tofile(buffer)
+
     def test_load_raises_on_path_traversal(self):
         tensor = _core.ExternalTensor(
             "../../etc/passwd",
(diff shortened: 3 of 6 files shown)


justinchuby:
@copilot please fix comments and lint

cyyever (approved):

titaiwangms:
## Review: Support sharding plain ONNX external data

Reviewed by a multi-model review team (readability, correctness/idiom, adversarial, and deep spec/math). **Overall the design is sound and the sharding math is provably correct** — the per-shard size estimate is alignment-aware and exactly matches the on-disk writer, the partition→write ordering invariant holds end-to-end (so the `zip()` that swaps in external tensors stays aligned), and the callback factory avoids the classic closure late-binding trap. Nicely tested. A few items below are worth addressing before merge.

### 🔴 Critical

**1. Re-saving a *loaded* sharded model to the same base can silently corrupt data.**
`convert_tensors_to_external()` only materializes-to-memory the input tensors that reference the **current destination file** (the `os.path.samefile(path, tensor.path)` check, `external_data.py:303-328`). In the shard loop, shards are written one at a time. If you load a sharded model (tensors are now `ExternalTensor`s pointing at the old shards) and re-save with a *different* shard layout, writing new shard 1 overwrites `model-00001-of-00002.data` while a tensor destined for new shard 2 may still point into that just-overwritten file — its bytes are read *after* being clobbered, producing corrupted output with no error.

Repro shape: save `[600,400,400]` with `max=1000` → shards `[600,400],[400]`; load; re-save with `max=800` → new shards `[600],[400,400]`. New shard 1 overwrites old shard 1, but a tensor in new shard 2 still references old shard 1.

Suggested fix: before writing any shard, materialize every input `ExternalTensor` whose path matches *any* target shard path (not just the current one); or write all shards to temp files and atomically rename after all source reads complete. Please also add a load→re-save-with-different-limit integrity test.

### 🟠 Major

**2. Shard filename generation corrupts relative paths with a dotted directory.** *(flagged by all four reviewers)*
`_get_shard_filename()` uses `base_name.rsplit(".", 1)` on the full path string (`external_data.py:137-146`). For e.g. `external_data="my.dir/model"` it produces `my-00001-of-00003.dir/model` — wrong directory and malformed name (likely `FileNotFoundError`). Fix: use `os.path.splitext()` (which only treats a dot in the final basename as an extension), splitting the dir component off first. Add a test with a dotted-directory / no-extension path.

**3. `max_shard_size_bytes` is documented as a hard "maximum" but isn't validated.**
`<= 0` is accepted and silently writes oversized shards. Validate early in `save()`/`unload_from_model()`: `raise ValueError` when `max_shard_size_bytes is not None and max_shard_size_bytes <= 0`.

**4. Stale shard files from a previous save are not cleaned up.**
Saving 5 shards then re-saving 2 to the same dir leaves `model-00003-of-00005.data` etc. The ONNX proto won't reference them, but `*.data` globs / directory uploads will leak old weights and bloat artifacts. Either clean up files matching the exact generated shard pattern for the chosen base name, or document that callers must save into a clean directory.

### 🟡 Minor

**5. A single tensor larger than `max_shard_size_bytes` is silently placed in an over-limit shard.** This is the correct unavoidable fallback, but emit a `logger.warning` (matching the existing style at `external_data.py:319`) and/or document the "best-effort for oversized tensors" contract.

**6. `_shard_tensors()` re-sorts the growing shard prefix on every append** (`_estimate_shard_size_bytes([*shards[-1], tensor])`), giving ~quadratic behavior with many small initializers. Consider incremental shard-size accounting.

**7. Readability cleanups:**
- Hoist `_make_shard_callback` out of the `for` loop (it's identical each iteration) — ideally to module level near the other helpers.
- Move `_estimate_shard_size_bytes` to immediately *before* `_shard_tensors` (its only caller) so it reads top-to-bottom.
- In `_shard_tensors`, `current_shard_size > 0` is an indirect proxy; `len(shards[-1]) > 0` expresses the real "don't open a new shard when current is empty" invariant and lets you drop the `current_shard_size` variable.
- Naming: `_get_shard_filename`'s `base_name` param actually receives a relative path; the `external_tensors_list` → `external_tensors` alias and the `shard_tensor_list` loop var can be simplified.

### Nit
- Under sharding, callback `index` reflects per-shard size-sorted write order, whereas the single-file path sorts globally. Indices are still unique/contiguous `0..N-1`; just worth a one-line doc note.

---
*Verified empirically by the deep reviewer (200 estimate-vs-actual mixes around the alignment threshold + 800 randomized sharding trials): estimate == actual file size, zero multi-tensor over-limit shards, partition order preserved. The only correctness gaps are the re-save path (#1) and the filename edge case (#2).*

justinchuby:
@copilot please address reviews

justinchuby:
@copilot+claude-opus-4.8 fix new lints and comments

titaiwangms:
## Re-review (round 2)

Thanks for the thorough revision — the round-1 items (dotted-path filenames, `<=0` validation, oversized-shard warning, O(1) sharding) are well handled. I independently **verified the new `_ShardSizeAccumulator` is mathematically correct**: a proof plus a 200,000-trial randomized fuzz (shuffled add-order vs the sorted reference) showed **zero divergences**, and the cross-shard slicing preserves initializer order so the final `zip()` stays aligned. 

However, the re-save safety fix and the new stale-cleanup introduce some serious issues — all reproduced locally.

### 🔴 Critical

**1. Re-saving a loaded sharded model with a *different* shard count crashes (`FileNotFoundError`).**
The round-1 re-save fix only covers the case where the new shard filenames *collide* with the old ones. When the shard **count** changes, old shard files have different names, so `_materialize_external_tensors_for_destination_paths()` (which only protects tensors matching a *new* destination path) doesn't load them — and then `_cleanup_stale_shard_files()` **deletes them before** `convert_tensors_to_external()` reads their data.

Reproduced: save 4 tensors with `max_shard_size_bytes=500` → 4 shards (`...-of-00004.data`); `ir.load`; re-save with `max_shard_size_bytes=900` → 2 shards →
```
FileNotFoundError: .../model-00001-of-00004.data  (in ExternalTensor.tofile)
```
The existing test `test_save_loaded_sharded_model_with_different_limit_preserves_data` misses this because `1000→800` keeps the count at 2→2 (filenames collide). Fix: defer cleanup until **after** all new shards are written, and/or materialize every input tensor whose backing file is slated for deletion — not just those matching a new destination path.

### 🟠 Major

**2. `_cleanup_stale_shard_files()` can delete *another* model's data in the same directory.**
Ownership is inferred from filename shape (`^{stem}-\d{5}-of-\d{5}{ext}$`) alone. Reproduced: a directory holding an unrelated `model-00001-of-00009.data` … `model-00009-of-00009.data` set — saving a *different* model as `external_data="model.data"` with `total_shards=2` **deleted all 9** of the other model's files. Cleanup should be scoped to files the previous save of *this* model actually referenced (e.g. from the prior proto's `location` fields), not a stem-based glob.

**3. Destructive cleanup is non-atomic and runs before the save is durable.**
`_cleanup_stale_shard_files()` runs inside `unload_from_model()` *before* the new shards and the `.onnx` proto are written. If any later step fails (callback raises, disk-full, permission error), old shards are already gone while the on-disk `.onnx` still references them → unrecoverable. Prefer: write new external data + proto successfully first (ideally temp files + atomic rename), then delete stale files.

### 🟡 Minor

**4. Cleanup is inconsistent across layout transitions.** Sharded→non-sharded (`max_shard_size_bytes=None`) never calls cleanup, leaving stale `...-of-NNNNN.data`. Single→multi leaves the old `model.data`. And `total_shards == 1` writes `model.data` but won't clean a prior `model-00001-of-00001.data`.

**5. Cleanup regex breaks above 99,999 shards.** `_get_shard_filename` uses `:05d` (minimum width), so `total_shards=100000` yields `...-of-100000.data`, which the cleanup regex `\d{5}` won't match. Either cap shards at 99,999 or use `\d{5,}`.

**6. Test gaps.** Add changed-shard-count re-save tests (2→3 and 3→2: assert no exception, data preserved, stale files removed). The accumulator boundary test builds tensors via float32 `nbytes // 4`, so `threshold + 1` collapses to `threshold` — the strict `> _ALIGN_THRESHOLD` boundary isn't actually exercised at +1; use exact-byte (`uint8`) tensors at `threshold-1/threshold/threshold+1`, and shuffle the accumulator's add order relative to the reference (its real usage is unsorted input).

### 🟢 Readability / Nits
- `_ShardSizeAccumulator.sum_aligned` actually holds *padded* bytes — rename to `sum_aligned_padded` to match `largest_aligned_padded/raw` and avoid a raw-vs-padded trap.
- Document the load-bearing invariant in `.size`: write order is ascending, so the largest aligned tensor is always written last and needs no trailing padding (which is why its padding is subtracted). Right now the formula reads as magic.
- Add one-line docstrings to `_materialize_external_tensors_for_destination_paths` (safety-critical: "load into memory any external tensor whose backing file is about to be overwritten") and `_cleanup_stale_shard_files`.
- `has_aligned` is redundant with `largest_aligned_raw > 0`.

---
*Verified: accumulator math (proof + 200k-trial fuzz, 0 divergences) and cross-shard ordering are correct. Findings #1 and #2 reproduced locally; tests still pass because neither scenario is covered. #1, #2, #3 share a root cause — eager destructive cleanup before durable writes + materialization that only covers new destination paths.*

src/onnx_ir/_io.py
```diff
@@ -92,17 +93,31 @@ def callback(tensor: ir.TensorProtocol, metadata: ir.external_data.CallbackInfo)
...
             Effective only when ``external_data`` is set.
+        max_shard_size_bytes: Maximum cumulative size in bytes for a single external data shard file.
+            When ``None`` (the default) all external tensors are written to the single file
+            given by ``external_data``. When set, tensors are distributed across numbered shard
+            files (e.g. ``model-00001-of-00003.data``). Because the ONNX format stores
+            ``location``, ``offset``, and ``length`` per tensor, no separate index file is
+            created — the saved ONNX proto itself encodes which shard each tensor lives in.
+            If a single tensor is larger than this value, it is written in its own oversized
+            shard file.
+            Effective only when ``external_data`` is set.
```
gramalingam, line 104: Should this be `requires ``external_data`` to be set`? Is there a warning/error otherwise? Or, is it ignored?

gramalingam:
Nit: the PR description "Safetensors sharding is incompatible with ORT CUDA kernels due to uncontrolled tensor alignment. " is a bit confusing. I am guessing it explains why an alternative implementation technique (safetensors sharding) wouldn't work? Better to start with *what* the PR does enable.

Another question: I assume users have the ability to explicitly manage externalization themselves? That is, in case they want to use more sophisticated bin-packing algorithms to map tensors into a smaller collection of external files? This provides one default simple implementation?

gramalingam:
Would be good for the API documentation to explicitly indicate behavior if files already exist: will they be overwritten, or is it a prerequisite that none of the files exist already (throw an error otherwise)? In the latter case, a check should probably be done first before writing anything (a little involved code probably to handle edge cases like file-creation between check and write).

src/onnx_ir/_io.py
justinchuby: I think it is currently ignored. I suppose we could raise an error? Do you have a preference?

justinchuby:
> Nit: the PR description "Safetensors sharding is incompatible with ORT CUDA kernels due to uncontrolled tensor alignment. " is a bit confusing. I am guessing it explains why an alternative implementation technique (safetensors sharding) wouldn't work? Better to start with _what_ the PR does enable.

Yes. Safetensors will have a json header before the data blob, which means the data may start at a non-aligned address. Additionally the format does not allow "holes" in the data region. So we cannot control alignment of each tensor. Will update PR description.

> Another question: I assume users have the ability to explicitly manage externalization themselves? That is, in case they want to use more sophisticated bin-packing algorithms to map tensors into a smaller collection of external files? This provides one default simple implementation?

Yes. users have the freedom to manipulate external data however they like. It is a simple matter of replacing the Tensor objects with ExternalTensor objects.

justinchuby:
Addressed the open review comments in 4339b02:

- **@gramalingam** ([_io.py:104](https://github.com/onnx/ir-py/pull/427#discussion_r3342893888)): `max_shard_size_bytes` without `external_data` was previously silently ignored. Now raises `ValueError` with a clear message; added `test_save_raise_when_max_shard_size_bytes_without_external_data`.
- **@Copilot** ([external_data.py:254](https://github.com/onnx/ir-py/pull/427#discussion_r3342197259), O(n²log n) re-sort): already fixed in 79c1b32 — `_shard_tensors` now uses `_ShardSizeAccumulator` which maintains shard size in O(1) per append via a few running aggregates (`sum_unaligned`, `sum_aligned`, `largest_aligned_*`). No sort per append. This thread can be resolved.

justinchuby:
@titaiwangms thanks for the thorough round-2 review. Pushed 35e75f4 addressing every finding:

**🔴 #1 (re-save with different shard count → `FileNotFoundError`)** — Materialization now covers every file the upcoming save will overwrite **or** delete (new destinations + stale shards), not just the immediate new destination. Regression test: `test_resave_loaded_model_with_changed_shard_count_preserves_data` (4 shards → 2).

**🟠 #2 (cross-model deletion)** — Cleanup is now ownership-scoped: a file is eligible for removal only if it was referenced by an `ExternalTensor` in the current model **or** by the prior on-disk `.onnx` proto at the same path. `_io.save` reads the existing `.onnx`'s `external_data` location entries via the new `_read_existing_external_data_locations` helper and passes them as `prior_external_files` into `unload_from_model`. Regression tests: `test_save_does_not_delete_unrelated_models_shards` + `test_cleanup_does_not_delete_unrelated_models_shards_in_same_directory`.

**🟠 #3 (non-atomic destructive cleanup)** — Cleanup runs **after** all new shards (and, in `_io.save`, after the `.onnx` proto) are written. A mid-save failure can no longer leave the model referencing already-deleted files. (Full temp-file + atomic-rename is a larger refactor; happy to do it as a follow-up if you'd prefer.)

**🟡 #4 (layout transitions)** — The non-sharded write path now also computes and removes stale shards, and the single-shard layout (`total_shards == 1`) cleans up any leftover `model-00001-of-00001.data`. Regression test: `test_resave_loaded_sharded_model_as_single_file_preserves_data`.

**🟡 #5 (>99,999 shards regex)** — Pattern changed from `\d{5}` to `\d{5,}` to match the natural-width formatter. Regression test: `test_shard_count_above_five_digits_uses_natural_width`.

**🟡 #6 (test gaps)** — Added uint8 boundary test (`test_incremental_size_at_strict_alignment_threshold_boundary`) so threshold±1 is actually exercised, plus shuffled-order fuzz (`test_incremental_size_invariant_under_add_order`).

**🟢 Nits** — `sum_aligned → sum_aligned_padded`; `has_aligned` removed (`largest_aligned_raw > 0` is the canonical predicate); expanded `.size` docstring with the leading-pad + last-largest-subtract invariant; added top-line docstrings to `_materialize_external_tensors_for_destination_paths`, `_find_stale_shard_files`, and `_delete_stale_shard_files`.

The old `test_cleanup_removes_invalid_zero_indexed_shard_files` was inverted into `test_cleanup_leaves_unowned_zero_indexed_shard_files_alone` to encode the new ownership contract. Full suite: **1626 passed, 8 subtests passed**.

justinchuby:
Simplified per @justinchuby's suggestion in b605494: dropped the ownership-scoped cleanup machinery (`_find_stale_shard_files`, `_delete_stale_shard_files`, the prior-proto reader, the two-phase materialize-then-delete ordering) in favor of a single rule —

> When saving sharded external data, refuse to clobber any existing destination file that is not already referenced by an `ExternalTensor` in the in-memory model. Raise `FileExistsError` instead.

**Net diff: −128 LOC** (`external_data.py` shrinks from ~370 lines of cleanup-related code to ~50 of just collision checking).

Behavior summary:
- **Sharded path**: any pre-existing destination file that isn't owned by the in-memory model → `FileExistsError` with a clear message ("Either delete these files first, save into a clean directory, or load the model with ir.load() before re-saving").
- **Single-file path**: unchanged, permissive overwrite. Collisions are much more dangerous under sharding (layouts are ambiguous), so the strict rule is sharded-only.
- **Self-overwrite** (`ir.load` → re-save same path): still works via the existing per-file materialization in `convert_tensors_to_external`.
- **No automatic cleanup**: stale shard files from a prior layout stay on disk — the caller is responsible. This trades a tiny convenience for completely eliminating the cross-model-deletion class of bug.

Tests rewritten to match the new contract:
- Positive: `test_save_sharded_does_not_raise_on_unrelated_shard_files`, `test_save_sharded_does_not_touch_unowned_zero_indexed_shard_files`, `test_save_sharded_does_not_delete_unrelated_models_shards_in_same_directory`.
- Negative: `test_save_sharded_raises_on_foreign_destination_collision` (unit-level), `test_save_sharded_raises_on_foreign_shard_collision` (via `_io.save`).
- The round-2 re-save tests still verify data round-tripping; the cleanup assertions are gone.

Full suite: **1628 passed**. Lint clean.

src/onnx_ir/_io.py
justinchuby: Raising an error instead

justinchuby:
@gramalingam @titaiwangms addressed comments. PTAL, thanks!

titaiwangms:
## Review — round 4 (collision-check redesign, HEAD `347e692`)

Thanks for the simplification — dropping the stale-shard cleanup machinery in favor of a strict foreign-collision check (`_check_no_foreign_shard_collisions`) makes the ownership model much easier to reason about, and the round-3 regressions stay fixed. The accumulator math and greedy-ordering invariant remain correct.

However, removing the global pre-materialization pass reopened a **data-corruption hole**. Our review team reproduced it independently three ways (two reviewers + a manual repro), so I'd block on it.

### 🔴 Critical — self-overwrite corruption when a tensor migrates to a *later* shard at constant shard count

`external_data.py:782` (shard loop) + `:573` (per-shard materialization).

Safety now rests entirely on the per-shard `_materialize_external_tensors_for_destination_paths(shard_tensors, [path])` call inside `convert_tensors_to_external`: when writing shard *k*, it only materializes tensors **in shard k's own list** that reference **shard k's** destination. It does **not** protect a tensor that still *backs* an earlier shard's file `F_i` but has been *reassigned to a later shard j > i* whose filename `F_j ≠ F_i`.

Because `_shard_tensors` is a greedy **contiguous** partition over initializer order, any in-memory edit that shifts a cut point — or even just re-saving with a different `max_shard_size_bytes` that yields the **same shard count** but a moved boundary — migrates a tensor across shards while keeping all filenames identical. The collision check then *passes* (the files genuinely are owned), shard *i* is rewritten and truncated **before** shard *j* reads the migrated tensor, and that tensor is never materialized by either shard's write. Result: silent corruption (or, when the stale offset lands past EOF, an infinite loop — see below).

Deterministic repro (load → resize one tensor → re-save to the same path, same `max_shard_size_bytes`, same shard count 2):

```python
import os, shutil, numpy as np, onnx_ir as ir
base = ".repro"; shutil.rmtree(base, ignore_errors=True); os.mkdir(base)
vals = [ir.Value(name=f"w{i}", const_value=ir.Tensor(np.arange(n, dtype=np.float32) + i*1000, name=f"w{i}"))
        for i, n in enumerate([75, 75, 75, 125])]  # bytes: 300,300,300,500
m = ir.Model(ir.Graph([], [], nodes=[], initializers=vals, name="g", opset_imports={"": 18}), ir_version=8)
p = os.path.join(base, "model.onnx")
ir.save(m, p, external_data="model.data", size_threshold_bytes=0, max_shard_size_bytes=900)  # [w0,w1,w2][w3]
m2 = ir.load(p)
# enlarge w1 600B -> shifts greedy cut to [w0,w1][w2,w3] (still 2 shards); w2 migrates shard1 -> shard2
m2.graph.initializers["w1"].const_value = ir.Tensor(np.arange(150, dtype=np.float32) + 10000, name="w1")
ir.save(m2, p, external_data="model.data", size_threshold_bytes=0, max_shard_size_bytes=900)
w2 = ir.load(p).graph.initializers["w2"].const_value.numpy()
print(np.array_equal(w2, np.arange(75, dtype=np.float32) + 2000))  # -> False; w2 now holds w1's bytes
```

Observed: `w2` reloads as `[10075.0 … 10149.0]` instead of `[2000.0 … 2074.0]` — a valid-looking ONNX model with wrong tensor contents, no error raised.

This is a realistic workflow (load a sharded model, edit weights/shapes, save back). It was *not* present in round 3, which materialized every tensor referencing **any** destination path before writing. The current tests pass only because they change the shard count (4→2, →single-file), so old/new filenames differ and nothing is overwritten in place — they structurally cannot catch the same-count case.

**Suggested fix:** in the sharded branch, after the collision check and before the shard loop, restore a global pre-materialization over **all** `tensors_to_externalize` against **all** `destination_paths` (the round-3 behavior, now scoped to just `destination_paths` since cleanup is gone). Alternatively, write shards to temp names and atomically rename so no live-referenced file is overwritten before all reads complete.

### 🟠 Major — `ExternalTensor.tofile` infinite-loops on a short/truncated read

`_core.py:914-917`. When the source file is shorter than `offset + length` (exactly what the bug above produces with `w1=400B`: new shard-1 is 700B but `w2` still points at offset 600 len 300), the copy loop reads 100 bytes, then `src.read()` returns `b""` at EOF, and `bytes_to_copy` is never decremented → spins forever. This is both an amplifier of the Critical and an independent DoS vector for any corrupt/truncated external-data file.

```python
while bytes_to_copy > 0:
    chunk = src.read(min(chunk_size, bytes_to_copy))
    if not chunk:
        raise OSError(f"External data file {self.path!r} is shorter than expected ...")
    file.write(chunk)
    bytes_to_copy -= len(chunk)
```

### 🟡 Minor

- **Missing regression test for the Critical.** Add a same-shard-count re-save test: load a 2-shard model, mutate an initializer (or re-save with a different `max_shard_size_bytes` yielding the same count) so a boundary shifts, save to the same base, reload and assert **all** tensors round-trip. This fails today and should gate the fix. (`_io_test.py`)
- **`_io.save` docstring** lists only `ValueError` under `Raises:` — add `FileExistsError` now that the sharded path can raise it. (`_io.py:108`)
- **Single-file vs sharded asymmetry** (pre-existing, not blocking): single-file external data still permissively overwrites a foreign `model.data` and leaves stale numbered shards on disk. Worth a follow-up to either align the policy or document it in the public docstring.
- **Owned-set gap:** `_collect_owned_external_files` only walks initializers; an external tensor referenced solely by a non-initializer value (e.g. a `Constant` attribute) sharing a shard base name would be treated as *foreign* and raise. Low likelihood; a comment or follow-up would help.

### 🔵 Readability nits

- `external_data.py:759` — the comment "the previous stale-shard cleanup machinery" references code deleted in this same PR; a fresh reader has no referent. Reframe to state the contract directly (refuse to clobber unowned files; different shard counts → different names → caller owns cleanup).
- `_check_no_foreign_shard_collisions` (`:407`) is generic (any paths + owned set), not shard-specific — consider `_check_no_foreign_file_collisions`.
- `_collect_owned_external_files` docstring says "normalized" but uses `os.path.realpath` — say "resolved (real) absolute paths" to avoid a wrong mental model.
- `unload_from_model` docstring doesn't document the single-file-vs-sharded raise asymmetry or the "stale shards are the caller's responsibility" contract — both currently live only in inline comments.

The collision-check idea is sound and correctly placed; it just needs to be paired with global materialization before any shard overwrite. Happy to re-review once the Critical is addressed.

justinchuby:
Thanks @titaiwangms — addressed round-4 in 3 commits (3f1c069, 4b15a02, a695258):

**🔴 Critical — same-shard-count boundary-shift data loss**: fixed by a single global pre-materialization pass over all `tensors_to_externalize` against all `destination_paths` before the shard write loop (a695258). Per-shard materialization was insufficient because a tensor migrating from shard `i` to shard `j > i` still pointed at file `i` while shard `j` was being written, and was lost once shard `i` got truncated. Added `test_resave_same_shard_count_with_boundary_shift_preserves_data` covering the `[75,75,75,125] → [75,150,75,125]` @ 900-byte cap repro you provided. I verified the new test fails on the prior tip and passes after the fix.

**🟠 Major — `ExternalTensor.tofile` infinite loop on short source**: `tofile` now raises `OSError` with the path, missing byte count, and the offset where the read stopped, instead of spinning forever on `b''` (4b15a02).

**🟡 Minors addressed**:
- Regression test added next to `test_resave_loaded_model_with_changed_shard_count_preserves_data`.
- `_io.save` docstring now lists `FileExistsError` under `Raises:`.
- Single-file vs sharded asymmetry: TODO left in both `_io.save` and `unload_from_model` docstrings (not fixed, as agreed).
- Owned-set gap: comment in `_collect_owned_external_files` body noting it only walks initializers passed in by callers and that this matches the writer's scope.

**🔵 Nits addressed**:
- The 'previous stale-shard cleanup machinery' comment near the shard loop now states the contract directly.
- Renamed `_check_no_foreign_shard_collisions → _check_no_foreign_file_collisions`.
- `_collect_owned_external_files` docstring now says 'resolved (real) absolute paths'.
- `unload_from_model` docstring documents the single-file/sharded raise asymmetry and the 'stale shards are caller's responsibility' contract.

All 1629 tests pass; lintrunner clean on the five touched files.

titaiwangms:
## Review — round 5 (fix for the round-4 Critical, HEAD `3f1c069`)

Thanks @justinchuby — the global pre-materialization fix is the right call, and our team (4 reviewers + manual repros) confirms it.

### ✅ The round-4 Critical is genuinely fixed
The single global pre-materialization pass over **all** `destination_paths` before the shard write loop (a695258) closes the same-shard-count boundary-shift corruption. Independently verified three ways:
- The original repro (`[75,75,75,125]` → enlarge w1 → same 2-shard count, w2 migrates shard1→shard2) now round-trips correctly.
- 2→3 (more shards), 3→2 (fewer shards, old `*-of-00003` left as documented), and multi-tensor reshuffle re-saves all preserve every initializer.
- Order/length alignment preserved 1:1 by `_materialize_external_tensors_for_destination_paths`; the per-shard materialization correctly degrades to a no-op afterward; the `tofile` short-read `OSError` offset math is exact.

Tests **547 passed**; ruff + mypy clean. The new `test_resave_same_shard_count_with_boundary_shift_preserves_data` is a genuine regression test (fails on the pre-fix tip, asserts all four initializers).

### 🟠 Major — legitimate load→edit→save can now falsely raise `FileExistsError`

`external_data.py` (`_collect_owned_external_files` / `_check_no_foreign_file_collisions`).

`owned_external_files` is inferred from the model's **surviving** `ExternalTensor` references, not from what the model was loaded from. So if you load a sharded model and **replace** an initializer that was the *sole owner* of a shard, the new in-memory tensor no longer references that shard file — and the collision check treats the model's **own** shard as foreign and refuses to save. Confirmed:

```python
# 3 tensors, each its own shard at cap 900; load, replace w1 (sole owner of shard 2), re-save same path
loaded.graph.initializers["w1"].const_value = ir.tensor(np.arange(200, dtype=np.float32) + 777, name="w1")
ir.save(loaded, p, external_data="model.data", size_threshold_bytes=0, max_shard_size_bytes=900)
# -> FileExistsError: Refusing to overwrite ... 'model-00002-of-00003.data'
```

`load → swap a weight → save in place` is a very common flow; today it fails unless the user manually deletes shard files. The cleanest fix is to write shards to temporary names and atomically rename into place — which also resolves both minors below.

### 🟡 Minor

- **Non-atomic self-overwrite.** If a shard write raises midway, truncated shard files are left on disk while the existing `.onnx` still references them, and the restored in-memory tensors are now invalidated. Pre-existing (not a regression from this fix), but worth a documented caveat or the temp-name+rename fix.
- **Self-overwrite RAM.** The global pass materializes every tensor backing a destination before writing, so an in-place re-save of a large model needs ~whole-model RAM — partly defeating sharding's memory goal. Inherent to the safe approach; temp-name+rename would let source shards stay on disk and avoid this.
- **Missing tests:** no unit test for the new `tofile` `OSError` short-read path (`_core_test.py`); no explicit few→many (2→3) same-base re-save test; no multi-tensor reshuffle re-save test (the boundary-shift test moves one key tensor).

### 🔵 Readability nits

- `_io.save` docstring splices `TODO(justinchuby):` mid-sentence into the rendered `Raises:` section — Sphinx/mkdocs will surface it to users. State the asymmetry declaratively in the docstring and move the tracking note to a code comment near the permissive single-file write path.
- The single-file/sharded asymmetry rationale is duplicated between the `unload_from_model` docstring and the inline comment; keep the "why" at the call site and trim the docstring.
- The "in-place edit shifts a shard boundary" comment reads as a file-level mutation — it means *editing the model's tensors* between saves (e.g. "a tensor was resized between saves").
- `tofile` `OSError`: extract `(self._offset or 0) + ((self._length or self.nbytes) - bytes_to_copy)` into a named `failed_at_offset` for readability on the error path.

### Praise
- The fix is the minimal correct one: a single global pass keyed on the complete destination set, consuming only shard *counts* from the pre-pass partition so there's no stale-object aliasing.
- `_materialize_external_tensors_for_destination_paths` uses `os.path.samefile` rather than string equality — correctly handles hard/symlinks pointing at a destination.
- The `tofile` short-read guard cleanly closes the infinite-loop failure mode with an actionable, offset-accurate message.
- The `_collect_owned_external_files` "Note" caveat about the initializers-only scope is exactly the kind of future-proofing that prevents a silent regression.

The only thing I'd treat as blocking is the Major (false `FileExistsError`); the temp-shard-names + atomic-rename approach would resolve it together with the two minors. Happy to re-review.

justinchuby:
I think it is ok to have the file exist error if the model is changed. Generally we can recommend users to save to a new, clean directory.

justinchuby:
@titaiwangms PTAL

titaiwangms (approved):
Thanks!

PR #437, opened 2026-06-09
Support multi-device configurations (object-bound)

justinchuby:
## Summary

Adds IRv11 multi-device configuration support to the IR, with an **object-bound** design: sharding specs and node device configurations bind directly to `Value` and `ModelConfiguration` objects rather than name strings. The proto `tensor_name` / `configuration_id` are *derived* from `value.name` / `configuration.name` at serialization time, giving a single source of truth and making references follow renames automatically.

This supersedes #428 (which used a name-based representation). It builds on that work and converts it to be object-bound, then adds a convenience API, an invariant checker, clone support, full tests, and user docs.

## What's included

- **Data model** (`_multi_device.py`): frozen dataclasses `ModelConfiguration`, `NodeDeviceConfiguration`, `ShardingSpec`, `ShardedDim`, `SimpleShardedDim`, `IndexToDeviceGroupMapEntry`. `ShardingSpec.value: Value` and `NodeDeviceConfiguration.configuration: ModelConfiguration` replace the name-string fields.
- **Serialization** (`serde.py`, `_multi_device.py`): names are derived on serialize; on deserialize, values resolve from node scope and configurations resolve via a model-level post-pass. Dangling references are preserved as lightweight placeholders so round-trips stay lossless. Serializing a sharding whose value has no name raises rather than emitting a malformed proto.
- **Convenience API** (`_core.py`): `Model.add_device_configuration(...)`, `Node.shard(value, configuration=…, axis=…, num_shards=…)` (validates eagerly), and `Node.sharding_of(value)`.
- **Validation**: an internal checker reports reachability, nameability, and structural problems (returns a list of messages; never raises for violations). It is not part of the public API yet. Serialization is the hard, fail-closed boundary.
- **Clone** (`_cloner.py`): sharding `Value` references are remapped through the clone's value map, so cloned models point at the cloned graph's values.

## Invariants

- **Single source of truth** — no second copy of names to drift out of sync.
- **References follow renames** — renaming a `Value` updates every sharding spec that points at it.
- **Edit freely, validate at checkpoints** — intermediate states may be temporarily invalid; validity is checked by the internal device-configuration checker and enforced hard (fail-closed) at the serialize boundary.

## Tests & docs

- `_multi_device.py` at **100% line/branch coverage**; covers serde edge cases, every checker branch, clone edge cases, convenience-method branches, and dangling-reference round-trips.
- New user guide `docs/multi_device.md` (with runnable examples) plus API reference entries. Docs build verified locally with zero new warnings.

## Verification

- Full test suite passes (excluding pre-existing torch-only tests in this environment).
- `ruff` (check + format) and `mypy` clean on changed files.


titaiwangms:
## Review synthesis — multi-device configurations (object-bound)

Reviewed by a five-model review team (readability, correctness, adversarial, spec-adherence, integration). The diff was applied and the suite run (**261 passed**), and edge cases were empirically round-tripped against onnx 1.22.0. No **Critical** issues. Strong overall design — full proto field coverage, dangling refs round-trip losslessly, the `dim=0` oneof footgun is dodged, and object identity is preserved across clone. Findings cluster around *fail-closed validation* and *graph-mutation ripple*.

### Major

**1. Checker doesn't enforce object identity — breaks the object-bound invariant**
`_multi_device.py:_check_device_configurations`. The checker matches `config.configuration.name in known_configs` but never requires the node's `ModelConfiguration` to *be* the registered object, and validates device indices using the node's own `num_devices`. A node bound to a same-named imposter config (`num_devices=100`) passes validation, then serializes to `conf0` and resolves to the real `num_devices=1` on load — silently invalid. Fix: error when `config.configuration is not known_configs[name]`; validate against the registered object's `num_devices`.

**2. Serialize emits partial/lossy metadata instead of failing closed**
`serde.py:_serialize_sharding_spec` / `serialize_node_device_configuration`. When `ShardingSpec.value is None` or `configuration is None`, serialization omits `tensor_name`/`configuration_id` rather than raising — yet the checker calls these invalid. Reachable via the cloner, which deliberately propagates dropped values to `ShardingSpec(value=None)`. Fix: reject `None` value/configuration at serialize; in the cloner/inliner, drop the affected spec (or the whole `NodeDeviceConfiguration`) when a value maps to `None`.

**3. `replace_input_with` / `resize_outputs` leave dangling `ShardingSpec` references**
`_core.py`. `Node.shard` enforces that a sharded value is an input/output of the node, but replacing an input or truncating outputs does not update/clean `node.device_configurations`. Result: specs point at dead `Value`s → invalid serialization or stale tensor names. Fix: in both methods, remap or drop `ShardingSpec` entries referencing the dropped value (mirroring `_cloner._remap_device_configurations`).

**4. `len(device) == num_devices` MUST not enforced; zero-device configs accepted**
The proto states `device` MUST be length `num_devices` if provided (`onnx.proto3:534`). `add_device_configuration(num_devices=4, device_names=("a","b"))` builds a config that fails `onnx.checker`, and the bare `add_device_configuration("conf")` yields `num_devices=0` — the same sentinel used for dangling-ref placeholders. Fix: enforce `len(device_names) == num_devices` when names are given, require `num_devices >= 1` for registered configs (or have the checker reject both).

### Minor

- **5. `pipeline_stage` sign-validation asymmetry**: `set_pipeline_stage` rejects `stage < 0`, but `Node.shard(pipeline_stage=...)` stores any int unchecked and serializes it. Pick one convention; add a guard + test in `shard`.
- **6. Silent drop when target proto lacks the field**: `hasattr` guards make `device_configurations` vanish without warning on older proto builds. Prefer a clear error/warning when IR metadata is present but the proto can't hold it.
- **7. `cascade=True` removal is identity-only**: `remove_device_configuration("conf0", cascade=True)` leaves same-named placeholder/manual configs on nodes. Document as identity-based, or also match by name when removal was requested by name.
- **8. `ModelConfiguration.device` (str names) vs `ShardingSpec.device` (int indices)** share a name with incompatible types; the docs even rename it to `device_names` in examples. Consider renaming the dataclass field to `device_names` to match the constructor kwarg.
- **9. Checker not wired into any serialize path** — only invoked from tests. Confirm whether validation should run automatically (ties into #1/#2).

### Nits (optional)

- `_check_device_configurations` is private + unexported, but the PR description advertises a public `ir.check_device_configurations(model)` and the docs name the private path — align naming/visibility/docs.
- `SymbolicDim(None)` normalizes to `dim=None` on round-trip — confirm intentional. All other `dim` cases (incl. `0`) round-trip exactly.
- Trim the "often negative" comment on group-map keys — not a proto rule.
- Naming consistency: `node_multi_device` → `device_configurations`; `_remap_` vs `_resolve_` verbs; a shared `_node_label` helper; the `INV-1..4` numbering has no definition site.
- Passes (`identity_elimination`, `constant_manipulation`, `replace_nodes_and_values`) silently drop `device_configurations` on node replacement; `__repr__`/`__str__`/`_display` omit it.

_Automated multi-agent review; please treat findings as suggestions to confirm, not blockers._

gramalingam (approved):

src/onnx_ir/_cloner.py
```diff
@@ -217,9 +219,47 @@ def clone_node(self, node: _core.Node, deep_copy: bool = False) -> _core.Node:
...
+        """Rewrite ``ShardingSpec.value`` references through the value map."""
+        if not device_configurations:
+            return device_configurations
+
+        new_configurations = []
+        changed = False
+        for configuration in device_configurations:
+            new_specs = []
+            spec_changed = False
+            for spec in configuration.sharding_spec:
```
justinchuby: sharding_specs should be plural

PR #440, opened 2026-06-11
Support sparse tensors

justinchuby:
## Summary

Adds first-class support for sparse tensors in the ONNX IR, including a new `SparseTensor` class (COO format following `SparseTensorProto`), serde support, sparse initializers, and sparse-tensor attributes.

This continues the work from #426 with a refined API.

### Sparse tensor support
- New `SparseTensor` class in `_core.py` storing `values`, `indices`, and `dims` (COO format), with `numpy()` (→ `scipy.sparse.coo_array`) and `as_tensor(lazy=)` (→ dense `Tensor`) conversions.
- `SparseTensor.from_scipy_sparse()` classmethod to construct from any `scipy.sparse` array (converted to COO). `ir.tensor()` stays dense-only with a clean `TensorProtocol` return type.
- `Value.const_sparse_value` field for sparse initializers, with correct name propagation and string representation.
- Serde: `serialize_sparse_tensor`/`serialize_sparse_tensor_into`/`deserialize_sparse_tensor`, sparse initializer (de)serialization, and `SPARSE_TENSOR`/`SPARSE_TENSORS` attribute handling.
- `SparseTensorProtocol` added to `_protocols.py`.

### Design notes
- `SparseTensor` intentionally has no `doc_string`/`metadata_props` because `SparseTensorProto` only stores `values`/`indices`/`dims`; `meta` is available for in-memory analysis (not serialized).

### Public API & docs
- Exposes `SparseTensor`, `AttrSparseTensor`, `AttrSparseTensors`, and the new serde functions.
- Documents sparse tensor usage in `docs/tensors.md`.

### Testing
- Comprehensive tests for construction, conversion, round-tripping, initializers, and attributes.

### DIFF
--- a/docs/tensors.md
+++ b/docs/tensors.md
@@ -79,7 +79,15 @@ Use {py:class}`onnx_ir.StringTensor <onnx_ir.StringTensor>` to create a string t
 
 ### Sparse Tensor
 
-Sparse tensors are not yet supported, but they are on our roadmap.
+Use {py:class}`onnx_ir.SparseTensor <onnx_ir.SparseTensor>` to create and work with sparse tensors in COO
+(coordinate) format, following the ONNX ``SparseTensorProto`` specification.
+
+A sparse tensor stores only the non-zero values and their indices, along with the dense shape. It can be:
+
+- Deserialized from an ``onnx.SparseTensorProto`` using {py:func}`onnx_ir.serde.deserialize_sparse_tensor <onnx_ir.serde.deserialize_sparse_tensor>`.
+- Serialized to an ``onnx.SparseTensorProto`` using {py:func}`onnx_ir.serde.serialize_sparse_tensor <onnx_ir.serde.serialize_sparse_tensor>`.
+- Used as a sparse initializer by setting ``value.const_value = sparse_tensor`` for a graph value registered as an initializer.
+- Used in node attributes via {py:func}`onnx_ir.AttrSparseTensor <onnx_ir.AttrSparseTensor>` and {py:func}`onnx_ir.AttrSparseTensors <onnx_ir.AttrSparseTensors>`.
 
 ## From `TensorProto`s and back
 
--- a/docs/tensors.md
+++ b/docs/tensors.md
@@ -84,9 +84,10 @@ Use {py:class}`onnx_ir.SparseTensor <onnx_ir.SparseTensor>` to create and work w
 
 A sparse tensor stores only the non-zero values and their indices, along with the dense shape. It can be:
 
+- Constructed from a ``scipy.sparse`` array using {py:meth}`onnx_ir.SparseTensor.from_scipy_sparse <onnx_ir.SparseTensor.from_scipy_sparse>`.
 - Deserialized from an ``onnx.SparseTensorProto`` using {py:func}`onnx_ir.serde.deserialize_sparse_tensor <onnx_ir.serde.deserialize_sparse_tensor>`.
 - Serialized to an ``onnx.SparseTensorProto`` using {py:func}`onnx_ir.serde.serialize_sparse_tensor <onnx_ir.serde.serialize_sparse_tensor>`.
-- Used as a sparse initializer by setting ``value.const_value = sparse_tensor`` for a graph value registered as an initializer.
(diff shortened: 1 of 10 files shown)


titaiwangms:
## Review summary — Support sparse tensors

Reviewed across readability, correctness, adversarial, spec-adherence, and cross-module integration lenses. One ship-blocking bug, confirmed independently with executable evidence (ONNX checker + numerical repros against `onnx.in.proto`).

### 🔴 Critical (blocking)

**2-D sparse indices are transposed relative to the ONNX spec.** The code uses `[rank, NNZ]`; `SparseTensorProto` requires `[NNZ, rank]` (each row is one nonzero's full coordinate tuple).
- `_core.py` `numpy()` / `as_tensor()`: `coords = tuple(indices_array[i] for i in range(indices_array.shape[0]))`
- `from_scipy_sparse()`: `np.stack(coords, axis=0)`

**Impact:** IR-created sparse tensors serialize to non-conformant protos; spec-compliant models with `NNZ > rank` **crash** (`IndexError: too many indices`), and asymmetric patterns silently densify to the wrong tensor. The test suite passes only because every 2-D case uses symmetric `NNZ == rank == 2`, where both layouts coincide.

**Fix:**
- read: `coords = tuple(indices_array[:, j] for j in range(indices_array.shape[1]))`
- write: `np.stack(coords, axis=1)`
- update docstrings (the "`[rank, NNZ]`" wording) and tests; add an asymmetric `NNZ != rank` round-trip test validated against ONNX's own densification.

### 🟠 Major

1. **`const_sparse_value` is not wired into existing consumers.** The field was added to `Value`, but many `if const_value is None` sites were not updated, so sparse initializers are silently dropped or crash:
   - `register_initializer` raises `ValueError` for sparse initializers (`_core.py`).
   - `_cloner.py` copies `const_value` but not `const_sparse_value` → cloning drops sparse data.
   - `_graph_containers.py` `tensors()` / `tensor_items()` / `get_tensor()` raise on sparse initializers.
   - `_safetensors/__init__.py` `assert tensor is not None` fails for sparse initializers.
   - `journaling/_wrappers.py` wraps `const_value.fset` but not `const_sparse_value.fset` → mutations bypass the journal.
   - `passes/common/_c_api_utils.py` strips sparse initializers before the C API call.
   - `passes/common/initializer_deduplication.py` logs false "invalid initializer" warnings.
2. **Deserialized sparse initializers are typed `SparseTensorType` instead of `TensorType`**, diverging from the dense initializer path and potentially altering model semantics (`serde.py`).
3. **String sparse tensors densify with `0` instead of `""`.** `np.zeros(dtype=object)` yields integer `0`; the spec default for string tensors is the empty string (`as_tensor`).
4. **No one-of invariant between `const_value` and `const_sparse_value`.** Both can be set simultaneously; serialization silently prefers sparse and a stale dense constant can be dropped. Consider clearing the sibling in each setter, or validating against `value.type`.
5. **Serialization mutates shared state.** `value.const_sparse_value.name = value.name` (and `SparseTensor.name` → `values.name` aliasing) mutates shared objects during serialization; unsafe when a tensor is shared across values/attributes. Prefer patching the proto after serialization.

### 🟡 Minor
- `as_tensor(lazy=True)` closure captures mutable `self`; snapshot `dims`/`values`/`indices`/`dtype` locally.
- `from_scipy_sparse` does not canonicalize indices; the spec requires ascending order without duplicates (`coo.sum_duplicates()` + sort).
- `values` shape docstring says `[nnz, ...]`; the spec defines `values` strictly as 1-D `[NNZ]`.
- `Value.__repr__` labels a sparse backing as `const_value={...}`; should read `const_sparse_value=`.
- Index-unpacking logic is duplicated verbatim in `numpy()` and `as_tensor()`; extract a single `_unpack_indices` helper.

### ✅ Praise
- Name ↔ `values.name` mapping is spec-faithful across the property, `Value.name` rename propagation, and serialization.
- 1-D linear-index path (`np.unravel_index`, C order) is correct.
- Sparse-vs-dense initializer routing and the targeted warning suppression are clean and well-tested.
- `SparseTensor` deliberately avoids structurally satisfying the runtime-checkable `TensorProtocol`, so `to_proto`/`from_proto` dispatch stays correct.

<sub>🤖 Generated with multi-model review (readability, code, critical, deep-spec, integration).</sub>

justinchuby:
@copilot please fix reviews

Issue #448, opened 2026-06-22
Double initializers with same name raise sorted error in NameFixPass


OYCN:
repro script
```
"""Minimal reproduction of onnx-ir NameFixPass bug.

When a graph contains a Value with const_value that is NOT registered in
graph.initializers but shares the same name as a registered initializer,
NameFixPass renames it to ensure uniqueness. However, it does NOT register
the renamed Value as an initializer. The result: the renamed Value becomes
a dangling reference (neither a graph input, nor an initializer, nor a node
output), causing a toposort validation failure on serialization.

Affected: onnx-ir 0.2.1, NameFixPass in onnx_ir.passes.common.naming

Expected behavior: NameFixPass should register the renamed Value as an
initializer (since it has const_value and was used as an input), or at
minimum the graph serializer should emit it.
"""

import logging

import numpy as np
import onnx
import onnx_ir as ir
from onnx_ir.passes.common.naming import NameFixPass

logging.getLogger("onnx_ir").setLevel(logging.ERROR)


def reproduce():
    # Build a simple model with one initializer named 'shape'
    model = ir.from_onnx_text(
        """\
<ir_version: 9, opset_import: ["" : 21]>
test (float[2, 3] x) => (float[6] out1, float[2, 3] out2) {
    out1 = Reshape (x, shape)
    out2 = Reshape (x, shape)
}""",
        initializers=[ir.Tensor(np.array([6], dtype=np.int64), name="shape")],
    )

    # Create a second Value with same name but different data, NOT in initializers.
    # This simulates what happens when a rewrite (e.g. ReshapeReshape) creates a
    # new initializer Value via op.initializer() but registration fails/is skipped.
    orphan = ir.Value(
        name="shape",
        shape=ir.Shape([2]),
        type=ir.TensorType(ir.DataType.INT64),
        const_value=ir.Tensor(np.array([2, 3], dtype=np.int64), name="shape"),
    )

    # Point first Reshape at the orphan value
    nodes = list(model.graph)
    nodes[0].replace_input_with(1, orphan)

    # Verify precondition: two Values named 'shape', only one registered
    assert nodes[0].inputs[1].name == "shape"
    assert nodes[0].inputs[1] is not model.graph.initializers["shape"]

    # Run NameFixPass - it renames orphan to 'shape_1' but doesn't register it
    NameFixPass()(model)

    # Show the result
    print("After NameFixPass:")
    for i, node in enumerate(model.graph):
        v = node.inputs[1]
[...]

PR #449, opened 2026-06-22
Fix NameFixPass not registering orphan initializers after rename

OYCN:
## Summary

> https://github.com/onnx/ir-py/issues/448
  
  When a `Value` with `const_value` (an orphan initializer) shares the same name as a registered initializer, `NameFixPass` renames it for uniqueness but
  does not register it in `graph.initializers`. This causes the value to become a dangling reference, leading to validation/serialization failures.
  
  ## Fix
  
  After renaming, register the value as an initializer if it has `const_value`, no producer, and is not already registered.
  
  ## Test
  
  Added `test_orphan_value_with_const_value_registered_as_initializer_after_rename` to verify the fix.

### DIFF
--- a/src/onnx_ir/passes/common/naming.py
+++ b/src/onnx_ir/passes/common/naming.py
@@ -167,6 +167,14 @@ class NameFixPass(ir.passes.InPlacePass):
                         input_value, scoped_used_value_names[-1], seen_values, value_counter
                     ):
                         modified = True
+                        # Register orphan values with const_value as initializers
+                        if (
+                            isinstance(graph_like, ir.Graph)
+                            and input_value.const_value is not None
+                            and input_value.producer() is None
+                            and not input_value._is_initializer
+                        ):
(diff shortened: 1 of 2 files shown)


justinchuby:
This is a bit tricky, as a value with const_value and not registered as initializer is not treated as an initializer by onnx ir. This is allowed because we also allow const_value to be used as a constant propagation field. So it's not always true that all values with const_value set are initializers.

With the bug you linked, I suppose we should properly register the value as an initialized value instead?

OYCN:
> This is a bit tricky, as a value with const_value and not registered as initializer is not treated as an initializer by onnx ir. This is allowed because we also allow const_value to be used as a constant propagation field. So it's not always true that all values with const_value set are initializers.
>
> With the bug you linked, I suppose we should properly register the value as an initialized value instead?

Got it. It seems this issue is caused by improper usage downstream, so I'll close this issue&PR and focus on the downstream side instead.

